PR #25, opened 2018-10-30
minimal working ros example

StefanJVA:


ros_ws/src/CMakeLists.txt
```diff
@@ -0,0 +1 @@
+/opt/ros/kinetic/share/catkin/cmake/toplevel.cmake
```
Dadie: Add the symlink to the .gitignore as the given path is not universal for each operating system

StefanJVA: Okay i am going to fix this

PR #35, opened 2018-11-01
Add autopep8 to the format script to automatically make code PEP8 com…

marian42:
…pliant

implements #16


scripts/format/format-src.sh
```diff
@@ -3,3 +3,4 @@ cd "$(dirname ${BASH_SOURCE[0]})"
 
 ./clang-format-script.sh ../../ros_ws/ file
 
+autopep8 --in-place --recursive --aggressive --aggressive ../../ros_ws/
```
Dadie: I would add a newline here as by unix convention (see https://robots.thoughtbot.com/no-newline-at-end-of-file)

PR #59, opened 2018-11-13
Clean and complete URDF description of the racer 

NklasF:
I added the package racer_model, which contains an URDF description of the racer. The main file is /racer_descritpion/racer.xacro. The racer has a camera and a hokuyo attached and they publish data to /racer/camera1/image_raw and /racer/laser/scan. Furthermore the racer can be driven through a differential drive, the relevent topic here is /cmd_vel. The racer can be viewed in Gazebo by launching launch/racer_gazebo.launch or launch/racer_rviz.launch for Rviz. Currently the racer is put in an empty wolrd.


ros_ws/src/racer_model/CMakeLists.txt
```diff
@@ -0,0 +1,195 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(racer_model)
+
+## Compile as C++11, supported in ROS Kinetic and newer
+# add_compile_options(-std=c++11)
```
StefanJVA:
```suggestion
add_compile_options(-std=c++11)
```

NklasF:
The Problem with the steering seems pretty odd indeed. But i can not see the problem.

PR #68, opened 2018-11-16
Sami remote controll

Sami19944:
- controller implementation
- keyboard implementation
- (partial) fail-safe implemtation


ros_ws/dev.launch
```diff
@@ -1,4 +1,40 @@
...
+    </node>
+
+    <!-- joy node -->
+    <node
+      respawn="true"
+      pkg="joy"
+      type="joy_node"
+      name="joy_node"
+      output="log" >
+      <param name="dev" type="string" value="/dev/input/js2" />
```
Nils-Jung: Why js2? The default should probably be js0, maybe it can even be an argument in the roslaunch command.
```suggestion
      <param name="dev" type="string" value="/dev/input/js0" />
```

ros_ws/src/auto_race_pg/include/remote_keyboard.h
```diff
@@ -0,0 +1,56 @@
...
+#include <std_msgs/Int64.h>
+
+#include <signal.h>
+#include <termios.h>
+#include <time.h>
+
+#define MODE "keyboard"
+
+#define TOPIC_SPEED "/set/speed"
+#define TOPIC_ANGLE "/set/position"
```
Nils-Jung: Shouldn't this be "/set/angle"?
```suggestion
#define TOPIC_ANGLE "/set/angle"
```

ros_ws/src/auto_race_pg/src/car_control.cpp
```diff
@@ -0,0 +1,96 @@
...
+{
+    if (!run)
+    {
+    }
+    else
+    {
+        // std::cout << "speed: " << speed << std::endl;
+        std_msgs::Float64 msg;
+        msg.data = speed * MAX_SPEED;
+        if (msg.data > MIN_SPEED)
```
Nils-Jung: Right now you cannot come to a full stop after driving forward since a speed of 0 is smaller than MIN_SPEED.

ros_ws/src/auto_race_pg/src/car_control.cpp
```diff
@@ -0,0 +1,96 @@
...
+    {
+    }
+    else
+    {
+        double            a = 0;
+        double            b = 1;
+        double            c = (1 - MAX_ANGLE);
+        double            d = MAX_ANGLE;
+        std_msgs::Float64 msg;
+        msg.data = ((d - b) * angle + (b * c - a * d)) / (c - a);
```
Nils-Jung: I really don't know what this formula is supposed to do. Are you trying to map the interval [a,b] to [c,d]? In that case, it should be:
```suggestion
        msg.data = (angle - a) / (b - a) * (d - c) + c;
```

ros_ws/src/auto_race_pg/include/remote_keyboard.h
```diff
@@ -0,0 +1,56 @@
...
+#define TOPIC_STATUS_DMS "/status/dms"
+
+#define KEYCODE_W 119
+#define KEYCODE_A 97
+#define KEYCODE_S 115
+#define KEYCODE_D 100
+#define KEYCODE_SPACE 32
+
+#define DELTA_SPEED_UP 0.2
+#define DELTA_SPEED_DOWN 0.1
```
Nils-Jung: I understand what DELTA_SPEED_UP is supposed to do (all changes in speed are in increments of DELTA_SPEED_UP) but what is the meaning of DELTA_SPEED_DOWN?

ros_ws/src/auto_race_pg/include/remote_keyboard.h
Sami19944: this is a parameter, will change it to js0

ros_ws/src/auto_race_pg/src/car_control.cpp
Sami19944: fixed. if the speed is smaller that MIN_SPEED, it will be set to 0

ros_ws/src/auto_race_pg/src/car_control.cpp
Sami19944: changed

ros_ws/src/auto_race_pg/include/remote_keyboard.h
Sami19944: if "W" is not pressed, the speed should go down by DELTA_SPEED_DOWN

Sami19944: has been removed all together

PR #81, opened 2018-11-30
Ignore VScode settings directory

marian42:


.gitignore
```diff
@@ -6,3 +6,5 @@ ros_ws/devel/
 ros_ws/src/CMakeLists.txt
 
 ros_ws/\.catkin_workspace
+
+.vscode/
```
Nils-Jung: I don't think this should be in the repo since it has nothing to do with the project. This is something every user should do themselves if they need it (in /ros.package/.git/info/exclude).

marian42:
@StefanJVA

PR #84, opened 2018-12-04
Stefan launch

StefanJVA:
Hey, i refactored/renamed a lot of stuff and created some more launch-files to make the whole project a bit more intuitive. IMO this had to be done eventually, so i figured "better sooner than later". I used c2c1d01283ecd077d73ae722579e880f5c1946b7 as the base, cause it already fixes a lot of stuff.

The new folder structure (inside the src-folder) is as follows:

- **car_control**: houses the code that is responsible for controlling the real car (listening for drive_params and passing them to the focbox)
- **teleoperation**: houses the code for keyboard and gamepad remote control
- **simulation**: houses everything related to gazebo and simulation (in contrast to the other folders, the simulation folder is not a ros-package by itself but contains ros-packages)
- **autonomous**: houses all code that is resposible for autonomous operation (Example: 1. it subscribes to sensor-topics, 2. makes all necessary calculations, 3. publishes drive_params). This package it empty right now.
- **drive_msgs**: contains necessary ros messages that are used by multiple other packages (right now only the drive_param.msg)

I hope u guys like it. If not, let me know in the review and we figure out a different way.

P.S. As this is a pretty substantial change, everyone should consider re-basing their existing branches ;)


ros_ws/src/autonomous/CMakeLists.txt
```diff
@@ -0,0 +1,200 @@
...
+#############
+
+## Add gtest based cpp test target and link libraries
+# catkin_add_gtest(${PROJECT_NAME}-test test/test_autonomous.cpp)
+# if(TARGET ${PROJECT_NAME}-test)
+#   target_link_libraries(${PROJECT_NAME}-test ${PROJECT_NAME})
+# endif()
+
+## Add folders to be run by python nosetests
+# catkin_add_nosetests(test)
```
marian42: This file has a lot of comments. I think they could be removed.

ros_ws/src/autonomous/package.xml
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0"?>
+<package format="2">
+  <name>autonomous</name>
+  <version>0.0.0</version>
+  <description>The autonomous package</description>
```
marian42: The package is not autonomous... It could be something like "This package contains all code related to autonomous operation of the car."

ros_ws/src/autonomous/package.xml
```diff
@@ -0,0 +1,65 @@
...
+  <!-- One maintainer tag required, multiple allowed, one person per tag -->
+  <!-- Example:  -->
+  <!-- <maintainer email="jane.doe@example.com">Jane Doe</maintainer> -->
+  <maintainer email="stefan@todo.todo">stefan</maintainer>
+
+
+  <!-- One license tag required, multiple allowed, one license per tag -->
+  <!-- Commonly used license strings: -->
+  <!--   BSD, MIT, Boost Software License, GPLv2, GPLv3, LGPLv2.1, LGPLv3 -->
+  <license>TODO</license>
```
marian42: This should reflect the licenses of the project.

ros_ws/src/autonomous/package.xml
```diff
@@ -0,0 +1,65 @@
...
+  <!-- Use build_export_depend for packages you need in order to build against this package: -->
+  <!--   <build_export_depend>message_generation</build_export_depend> -->
+  <!-- Use buildtool_depend for build tool packages: -->
+  <!--   <buildtool_depend>catkin</buildtool_depend> -->
+  <!-- Use exec_depend for packages you need at runtime: -->
+  <!--   <exec_depend>message_runtime</exec_depend> -->
+  <!-- Use test_depend for packages you need only for testing: -->
+  <!--   <test_depend>gtest</test_depend> -->
+  <!-- Use doc_depend for packages you need only for building documentation: -->
+  <!--   <doc_depend>doxygen</doc_depend> -->
```
marian42: These comments should be taken out IMO.

ros_ws/src/teleoperation/src/remote_joy.cpp
```diff
@@ -6,8 +6,8 @@
  */
 RemoteJoy::RemoteJoy()
 {
-    out_speed = nh_.advertise< std_msgs::Float64 >(TOPIC_SPEED, 1);
-    out_angle = nh_.advertise< std_msgs::Float64 >(TOPIC_ANGLE, 1);
+    out_drive_param =
```
marian42: Please use "parameters" instead of "param" for better readability.

marian42: Don't change it as I'm making a new PR where I rename it.

ros_ws/src/autonomous/CMakeLists.txt
StefanJVA: All of these comments are auto-generated by catkin. So for every new ros package catkin fills up the CMakeLists.txt and package.xml with a lot of comments. In theory these should help with configuration. 
But i agree that they also make those files much harder to read. Therefore i suggest that we remove all of those comments in all of our ros packages.

PR #89, opened 2018-12-07
Improve code reability and naming

marian42:
- Makes some names more descriptive
- Make naming consistent (also follows ROS style guide)
- Remove some unused code

Note: it appears like a lot of code was added, but most of it was only moved and slightly changed.


ros_ws/src/car_control/src/car_controller.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+void CarController::publishDriveParameters(double raw_speed, double raw_angle)
+{
+    double speed = std::max(0.0, raw_speed * MAX_SPEED);
+    double angle = (raw_angle * MAX_ANGLE + 1) / 2;
+
+    if (this->enabled)
+    {
+        std_msgs::Float64 speed_message;
+        speed_message.data = speed;
```
Dadie: in my experience, it can help to use scoping to emphasise that an object/list/whatever is filled with data.
Here you could write:
```c++
std_msgs::Float64 speed_message;
{
    speed_message.data = speed;
}
```
instead of 
```C++
std_msgs::Float64 speed_message;
speed_message.data = speed;
```

ros_ws/src/car_control/src/car_controller.cpp
```diff
@@ -0,0 +1,67 @@
...
+        this->enabled = true;
+    }
+}
+
+int main(int argc, char** argv)
+{
+    ros::init(argc, argv, "car_controller");
+    CarController carController;
+    ros::spin();
+    return 0;
```
Dadie: May replace the magical number `0` with the Macro `EXIT_SUCCESS` from `<cstdlib>`

ros_ws/src/teleoperation/include/keyboard_controller.h
```diff
@@ -0,0 +1,34 @@
...
+#include <termios.h>
+#include <time.h>
+
+#define TOPIC_DRIVE_PARAMETERS "/set/drive_param"
+
+#define KEYCODE_W 119
+#define KEYCODE_A 97
+#define KEYCODE_S 115
+#define KEYCODE_D 100
+#define KEYCODE_SPACE 32
```
Dadie: May replace the Macros with an enum class e.g.
```C++
enum class keycode : int {
    A = 119,
    W = 97,
    S = 115,
    D = 100,
    SPACE = 32
};
```

ros_ws/src/teleoperation/src/keyboard_controller.cpp
```diff
@@ -0,0 +1,87 @@
...
+    std::cout << "Listening to keyboard..." << std::endl;
+    std::cout << "========================" << std::endl;
+
+    double velocity = 0;
+    double angle    = 0;
+    while (ros::ok())
+    {
+        int key = this->getKeyboardCharacter();
+
+        if (key == KEYCODE_W)
```
Dadie: May replace the `if`s with a single `switch(key){ case A: break; case B: break;}`

ros_ws/src/teleoperation/src/keyboard_controller.cpp
```diff
@@ -0,0 +1,87 @@
...
+    // back up current terminal settings
+    tcgetattr(STDIN_FILENO, &old_terminal);
+    new_terminal = old_terminal;
+    // disable buffering
+    new_terminal.c_lflag &= ~ICANON;
+    // apply new settings
+    tcsetattr(STDIN_FILENO, TCSANOW, &new_terminal);
+
+    // read character (non-blocking)
+    int character = getchar();
```
Dadie: is the the `<cstdio>` `getchar()`? Because the `<cstdio>` blocks on input (AFAIK).

marian42: This will be fixed in #93.

PR #99, opened 2018-12-11
racecar simulation from python to c++

FlyingdutchmanC:
and Ackermann steering implemented to set different angles for front wheeels


ros_ws/src/simulation/racer_control/include/drive_param_converter.h
```diff
@@ -0,0 +1,102 @@
...
+    
+
+    private:
+    /**
+     * @brief 
+     * 
+     * @param angle Angle of the center of the front axis.
+     * @return Angles One Ackermann angle for each front wheel
+     */
+    Angles calculateSteeringAngles(const double& angle);
```
Dadie: The const ref angle seems unnecessary complex. In worst case it adds a unnecessary indirection to the function call.

ros_ws/src/simulation/racer_control/include/drive_param_converter.h
```diff
@@ -0,0 +1,102 @@
+#pragma once
+#include <string>
+#include <cmath>
```
Dadie: <cmath> isn't used in the header file. Move this include to the source file could increase compile time.

ros_ws/src/simulation/racer_control/src/drive_param_converter.cpp
```diff
@@ -0,0 +1,76 @@
...
+}
+
+/**
+ * @brief 
+ * 
+ * @param argc 
+ * @param argv 
+ * @return int 
+ */
+int main(int argc, char** argv)
```
Dadie: You could put the main-function in it's own file to separate the class from the program.

ros_ws/src/simulation/racer_control/include/drive_param_converter.h
```diff
@@ -0,0 +1,102 @@
...
+     */
+    double m_length;
+
+    /**
+     * @brief 
+     * 
+     */
+    double m_width;
+
+    std::string m_TOPIC_DRIVE_PARAMETERS;
```
Dadie: See `m_length`

Issue #93, opened 2018-12-07
Improve keyboard controller


marian42:
- Make it possible to handle multiple keys pressed
- Make keyboard driving more smooth

PR #105, opened 2018-12-14
Rewrite keyboard controller so that concurrent key presses are handle…

marian42:
…d and driving is smoother

implements #93

### DIFF
--- a/ros_ws/src/teleoperation/include/keyboard_controller.h
+++ b/ros_ws/src/teleoperation/include/keyboard_controller.h
@@ -28,9 +28,9 @@ enum class KeyIndex : int
     STEER_RIGHT = 3
 };
 
-constexpr std::array<Keycode, 4> KEY_CODES = { Keycode::W, Keycode::A, Keycode::S, Keycode::D };
-
 constexpr int KEY_COUNT = 4;
+
+constexpr std::array<Keycode, KEY_COUNT> KEY_CODES = { Keycode::W, Keycode::A, Keycode::S, Keycode::D };
 
 constexpr double PARAMETER_UPDATE_FREQUENCY = 90;
 
@@ -46,7 +46,7 @@ constexpr double MAX_STEERING = 0.7;
 // MAX_STEERING is multiplied by this when travelling at MAX_VELOCITY, by 1.0 when resting and by an interpolated value
 // otherwise
 constexpr double FAST_STEER_LIMIT = 0.6;
-// The velocity clamped so that its absolute value is not greater than this
+// The velocity is clamped so that its absolute value is not greater than this
 constexpr double MAX_VELOCITY = 8;
 
 // When no steering key is pressed, the steering value will change towards 0 at this rate, in units per second
@@ -69,7 +69,7 @@ class KeyboardController
 
     SDL_Window* m_window;
 
-    std::array<bool, 4> m_key_pressed_state = { { false, false, false, false } };
+    std::array<bool, KEY_COUNT> m_key_pressed_state = { { false, false, false, false } };
 
     double m_velocity = 0;
     double m_angle = 0;
--- a/ros_ws/src/teleoperation/src/keyboard_controller.cpp
+++ b/ros_ws/src/teleoperation/src/keyboard_controller.cpp
@@ -17,6 +17,10 @@ double map(double value, double in_lower, double in_upper, double out_lower, dou
  * */
 KeyboardController::KeyboardController()
 {
+    ROS_ASSERT_MSG(KEY_CODES.size() == KEY_COUNT, "KEY_CODES needs to have KEY_COUNT many elements.");
+    ROS_ASSERT_MSG(this->m_key_pressed_state.size() == KEY_COUNT,
+                   "m_key_pressed_state needs to have KEY_COUNT many elements.");
+
     this->m_drive_parameters_publisher =
         this->m_node_handle.advertise<drive_msgs::drive_param>(TOPIC_DRIVE_PARAMETERS, 1);
     this->createWindow();


ros_ws/src/teleoperation/include/keyboard_controller.h
```diff
@@ -1,37 +1,82 @@
...
 
+enum class KeyIndex : int
+{
+    ACCELERATE = 0,
+    DECELERATE = 2,
+    STEER_LEFT = 1,
+    STEER_RIGHT = 3
+};
+
+constexpr Keycode KEY_CODES[] = { Keycode::W, Keycode::A, Keycode::S, Keycode::D };
```
Dadie: This c-style array could be replaced by a `std::array`. This would allow "foreach"-loop e.g.
```C++
for(auto key : KEY_CODE){/* DO STUFF */}
```
which might me useful in the implementation of the `pollKeyboardEvents` member function.

ros_ws/src/teleoperation/include/keyboard_controller.h
```diff
@@ -1,37 +1,82 @@
...
+// When no steering key is pressed, the steering value will change towards 0 at this rate, in units per second
+constexpr double STEERING_GRAVITY = 2;
+// When no throttle key is pressed, the velocity will change towards 0 at this rate, in units per second
+constexpr double THROTTLE_GRAVITY = 3;
+
 class KeyboardController
 {
     public:
     KeyboardController();
+    ~KeyboardController();
```
Dadie: By implementing a destructor, a default copy/move constructor is not auto generated by the compiler. Please explicit add  code to declare either the copy and move constructor as `delete` or `default`
e.g.
```C++
// Move
KeyboardController ( KeyboardController && ) = default;
// Copy
KeyboardController ( const KeyboardController & ) = default;
```

ros_ws/src/teleoperation/src/keyboard_controller.cpp
```diff
@@ -1,73 +1,126 @@
 #include "keyboard_controller.h"
 
+double clamp(double value, double lower, double upper)
+{
+    return min(upper, max(value, lower));
```
Dadie: Please add `#include <algorithm>`

ros_ws/src/teleoperation/src/keyboard_controller.cpp
```diff
@@ -1,73 +1,126 @@
...
-                break;
-            case Keycode::A:
-                angle += 1;
-                break;
-            case Keycode::D:
-                angle -= 1;
-                break;
+            for (int i = 0; i < KEY_COUNT; i++)
+            {
+                if ((uint16_t)event.key.keysym.sym == (int)KEY_CODES[i])
```
Dadie: Here auto-casting might results in a undesired result. Either cast both values to the same type or declare `KEY_CODES` the same type as `event.key.keysym.sym`.

ros_ws/src/teleoperation/src/keyboard_controller.cpp
```diff
@@ -1,73 +1,126 @@
...
 }
 
 void quitSignalHandler(int signal)
 {
     ros::shutdown();
-    exit(0);
+    exit(EXIT_SUCCESS);
 }
 
 int main(int argc, char** argv)
```
Dadie: May also separate the class implementation from the program.

ros_ws/src/teleoperation/include/keyboard_controller.h
```diff
@@ -1,37 +1,82 @@
...
 
     private:
-    ros::NodeHandle node_handle;
+    ros::NodeHandle m_node_handle;
+
+    ros::Publisher m_drive_parameters_publisher;
+
+    SDL_Surface* m_window;
+
+    bool m_key_pressed_state[4] = { 0, 0, 0, 0 };
```
Dadie: Please replace the 0's with the bool literal `false`
Might also replace the c-style array with a std::array.

ros_ws/src/teleoperation/src/keyboard_controller.cpp
```diff
@@ -1,73 +1,126 @@
...
 
-    double velocity = 0;
-    double angle = 0;
-    while (ros::ok())
+void KeyboardController::createWindow()
+{
+    if (SDL_Init(SDL_INIT_VIDEO) < 0)
     {
-        auto key = static_cast<Keycode>(this->getKeyboardCharacter());
+        throw std::runtime_error("Could not initialize SDL");
```
Dadie: Please add `#include <stdexcept>`

ros_ws/src/teleoperation/src/keyboard_controller.cpp
```diff
@@ -1,73 +1,126 @@
...
-    tcsetattr(STDIN_FILENO, TCSANOW, &new_terminal);
-
-    // read character (non-blocking)
-    int character = getchar();
-
-    // restore old settings
-    tcsetattr(STDIN_FILENO, TCSANOW, &old_terminal);
-
-    return character;
+    double steer = this->m_key_pressed_state[(int)KeyIndex::STEER_LEFT]
```
Dadie: Replace the `int`-cast with an `size_t` cast.

ros_ws/src/teleoperation/src/keyboard_controller.cpp
```diff
@@ -1,73 +1,126 @@
...
+
+    double steer_limit = map(fabs(this->m_velocity), 0, MAX_VELOCITY, 1, FAST_STEER_LIMIT);
+    double angle_update = steer * delta_time * STEERING_SPEED;
+    this->m_angle = clamp(this->m_angle + angle_update, -MAX_STEERING * steer_limit, +MAX_STEERING * steer_limit);
+    double velocity_update = throttle * delta_time * (this->m_velocity * throttle > 0 ? ACCELERATION : BRAKING);
+    this->m_velocity = clamp(this->m_velocity + velocity_update, -MAX_VELOCITY, +MAX_VELOCITY);
+
+    if (steer == 0 && this->m_angle != 0)
+    {
+        double sign = this->m_angle > 0 ? 1 : -1;
```
Dadie: You could replace this with 
```C++
auto sign = copysign(1.0, m_angle);
```
See https://en.cppreference.com/w/cpp/numeric/math/copysign for more information

ros_ws/src/teleoperation/src/keyboard_controller.cpp
```diff
@@ -1,73 +1,126 @@
...
+        {
+            this->m_angle = 0;
+        }
+    }
+
+    if (throttle == 0 && this->m_velocity != 0)
+    {
+        double sign = this->m_velocity > 0 ? 1 : -1;
+        this->m_velocity -= THROTTLE_GRAVITY * delta_time * sign;
+        if (fabs(this->m_velocity) < THROTTLE_GRAVITY * delta_time)
```
Dadie: either add fabs to the current namespace (`using std::fabs` or add `std::` to `fabs`. You might as well replace `fabs` with `abs` to increase the readability of the code (as some might be confused of the letter f thinking the value is casted to an float).

ros_ws/src/teleoperation/include/keyboard_controller.h
```diff
@@ -1,37 +1,82 @@
 #pragma once
 
-#include <ros/ros.h>
-
+#include <SDL.h>
 #include <drive_msgs/drive_param.h>
-#include <std_msgs/Int64.h>
-
+#include <ros/ros.h>
```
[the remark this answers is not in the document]
marian42: This is enforced by the autoformatter.

ros_ws/src/teleoperation/include/keyboard_controller.h
```diff
@@ -1,37 +1,87 @@
...
+{
+    ACCELERATE = 0,
+    DECELERATE = 2,
+    STEER_LEFT = 1,
+    STEER_RIGHT = 3
+};
+
+constexpr std::array<Keycode, 4> KEY_CODES = { Keycode::W, Keycode::A, Keycode::S, Keycode::D };
+
+constexpr int KEY_COUNT = 4;
```
Dadie: You could move the `contexpr int KEY_COUNT` above the  `constexpr std::array<Keycode, 4>` and use the `constexpr` instead of the number. `KEY_COUNT` could also be used in ` std::array<bool, 4> m_key_pressed_state` instead of the `4`.

ros_ws/src/teleoperation/src/keyboard_controller.cpp
```diff
@@ -1,84 +1,149 @@
...
-            case Keycode::A:
-                angle += 1;
-                break;
-            case Keycode::D:
-                angle -= 1;
-                break;
-            // Ignore other Keys
-            default:
-                break;
+            for (size_t i = 0; i < KEY_COUNT; i++)
```
Dadie: You could add an assertion here to make sure, `KEY_CODES` and `m_key_pressed_state` have the same size. Because both are arrays a [static assertion](https://en.cppreference.com/w/cpp/language/static_assert) should be possible e.g.
```C++
static_assert(KEY_CODES.size() == m_key_pressed_state.size(), "KEY_CODES and m_key_pressed_state needs to be of the same size");
```

PR #119, opened 2018-12-18
Wall following alper malik

DigitalWayfarer:
Car is now able to follow the left wall in addition to the right wall. It can only follow ONE wall at the time, either the left one or the right one. The wall to be followed can be changed in the wall_following.cpp in the callback method. The ability to follow the centerline between TWO walls still has to be implemented. Code was also cleaned up.

Marian's race track fix has to be merged into the master too. Otherwise the car will get stuck in the invisible wall just before the end of the lap.

### DIFF
--- a/ros_ws/src/autonomous/src/wall_following.cpp
+++ b/ros_ws/src/autonomous/src/wall_following.cpp
@@ -1,0 +1,329 @@
+#include <array>
+#include <cmath>
+#include <cstdlib>
+#include <iostream>
+
+#include "drive_msgs/drive_param.h"
+#include "drive_msgs/pid_input.h"
+#include "sensor_msgs/LaserScan.h"
+#include "std_msgs/Float64.h"
+#include <ros/console.h>
+#include <ros/ros.h>
+
+float g_right_error = 0;
+float g_right_prev_error = 0;
+float g_right_corrected_angle = 0;
+
+float g_left_error = 0;
+float g_left_prev_error = 0;
+float g_left_corrected_angle = 0;
+
+class WallFollowing
+{
+    public:
+    WallFollowing();
+
+    private:
+    ros::NodeHandle m_node_handle;
+
+    ros::Subscriber lidar_subscriber;
+    void lidarCallback(const sensor_msgs::LaserScan::ConstPtr& lidar);
+
+    ros::Publisher pid_publisher;
+
+    // void adjustSpeed(double speed);
+};
+
+WallFollowing::WallFollowing()
+{
+    lidar_subscriber =
+        m_node_handle.subscribe<sensor_msgs::LaserScan>("/racer/laser/scan", 1, &WallFollowing::lidarCallback, this);
+    pid_publisher = m_node_handle.advertise<drive_msgs::pid_input>("/pid_input", 1);
+}
+
+/**
+ * @brief Using the lidar scans and a given angle theta
+ * this method returns the range at the given theta after doing boundary checks.
+ *
+ * @param lidar
+ * @param theta
+ * @return float
+ */
+float getRange(const sensor_msgs::LaserScan::ConstPtr& lidar, float theta)
+{
+    // check if calculated index is invalid
+    // if so then return the chosen max range (here 4)
+    if (theta / 0.375f < 0 || theta / 0.375f > 719)
+    {
+        return 4;
+    }
+
+    // ranges size is 720, angle range is 270°
+    // so one step equals an angle of 0,375°
+    // to access range at angle theta, you have to divide the given theta by the step size
+    float range = lidar->ranges[theta / 0.375];
+
+    // check for absurd or nan values
+    // the max range should be 4 so the car doesn't get affected by walls very far away
+    if (range > 4 || std::isinf(range))
+    {
+        range = 4;
+    }
+    return range;
+}
+
+/**
+ * @brief Checks if there is an unavoidable wall in front of the car using the lidar scans
+ * and returns true to inhibit driving if it is the case.
+ *
+ * @param lidar
+ * @return bool
+ */
+bool emergencyStop(const sensor_msgs::LaserScan::ConstPtr& lidar)
+{
+
+    // used for range averaging
+    float front_range_sum = 0;
+
+    // calculate the average distance in front of the car (6° radius)
+    // used for robustness
+    // (e.g. there is noise and only a single index shows a close range...
+    // ... this is probably not an obstacle)
+    for (int i = 133; i < 139; i++)
+    {
+        front_range_sum += getRange(lidar, i);
+        // ROS_INFO_STREAM("front range sum: " << front_range_sum);
+    }
+
+    // return 0 (stop) if the object is too close
+    if ((front_range_sum / 6) < 0.3)
+    {
+        ROS_INFO_STREAM("too close!: " << front_range_sum);
+        return true;
+    }
+    else
+    {
+        return false;
+    }
+}
+
+/**
+ * @brief Using the lidar scans this method calculates the necesarry angle and velocity
+ * for the car to stay parallel to the right wall and then returns them both in that order.
+ *
+ * @param lidar
+ * @return std::array<float, 2>
+ */
+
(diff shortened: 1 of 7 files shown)


ros_ws/src/autonomous/src/wall_following.cpp
```diff
@@ -0,0 +1,336 @@
...
+#include <math.h>
+
+
+float g_right_error = 0;
+float g_right_prev_error = 0;
+float g_right_corrected_angle = 0;
+
+float g_left_error = 0;
+float g_left_prev_error = 0;
+float g_left_corrected_angle = 0;
```
Dadie: May put those global variables at least in a `struct` e.g. 
```C++
template<typename T>
struct control_correction_data {
    T error;
    T prev_error;
    T corrected_angle;
};
control_correction_data<float> g_right;
control_correction_data<float> g_left;
```
the `template` would allow to easily change the type to `double` or another type later on.

Also according to a [stackoverflow post](http://web.archive.org/web/20181219233245/https://stackoverflow.com/questions/37502096/ros-binding-a-callback-function-and-a-object-member-to-a-subscriber-node) you should be able to put all the global variables in the class `WallFollowing` without any problems.

ros_ws/src/autonomous/src/wall_following.cpp
```diff
@@ -0,0 +1,336 @@
...
+ * @param lidar 
+ * @param theta 
+ * @return float 
+ */
+float getRange(const sensor_msgs::LaserScan::ConstPtr& lidar, float theta)
+{	
+	// ranges size is 720, angle range is 270°
+	// so one step equals an angle of 0,375°
+	// to access range at angle theta, you have to divide the given theta by the step size
+	float range = lidar->ranges[theta/0.375];
```
Dadie: Please check before accessing the array if the calculated index is valid e.g.
```C++
class WallFollowing
{
    /* ... */
    public:
    using float DEFAULT_RANGE = 4;
    /* ... */
}
/* ... */
if (theta/0.375f < lidar->range_min || theta/0.375f > lidar->range_max) {
    return WallFollowing::DEFAULT_RANGE;
}
/* ... */
float range = lidar->ranges[theta/0.375f];
```

ros_ws/src/autonomous/src/wall_following.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+	// check for absurd or nan values
+	if(range > 4 || std::isinf(range))
```
Dadie: Why is anything above the magic number 4 absurd?

ros_ws/src/autonomous/src/wall_following.cpp
```diff
@@ -0,0 +1,336 @@
...
+	{
+		range = 4;
```
Dadie: Why 4?

ros_ws/src/autonomous/src/wall_following.cpp
```diff
@@ -0,0 +1,336 @@
...
+}
+
+/**
+ * @brief Checks if there is an unavoidable wall in front of the car using the lidar scans
+ * and returns a signal (0) to inhibit driving if it is the case.
+ * 
+ * @param lidar 
+ * @return float 
+ */
+float emergencyStop(const sensor_msgs::LaserScan::ConstPtr& lidar)
```
Dadie: Why return float? it seems like you only return 1 or 0 and only check whether or not the return value was not 0, might as well switch to `bool`.

ros_ws/src/autonomous/src/wall_following.cpp
```diff
@@ -0,0 +1,336 @@
...
+	// the lidar has a range of 270°, the algorithm use 180°
+	// so the the angle 0° for the algorith starts at angle 45° for the lidar
+	float a = getRange(lidar,theta);
+	float b = getRange(lidar,45);
+
+	ROS_INFO_STREAM("right a: " << a);
+	ROS_INFO_STREAM("right b: " << b);
+
+	// transform the difference between the angles at a and b into radians
+	float swing = (fabs(theta - 45) * M_PI) / 180;
```
Dadie: May replace the call of `fabs` with the namespaced version `std::fabs` or `std::abs` with the later having the advantage of not (even though in this case rightly) giving anyone the assumption the number given or returned has to be of type `float`.

ros_ws/src/autonomous/src/wall_following.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+	// alpha is the orientation of the car with respect to the wall
+	// and a_b is the distance to the wall
+	// a_c is the distance between the current position of the car and the predicted position
+	// we use the car_length as a_c for now PLEASE CHANGE IN THE FUTURE
+	// c_d is the distance to the wall at the predicted position
+	float alpha = atan((a*cos(swing)-b)/(a*sin(swing)));
+	float a_b = b*cos(alpha);
+	float a_c = 0.5;
+	float c_d = a_b + a_c*sin(alpha);
```
Dadie: May add `std::` to `atan`, `cos` and `sin`

ros_ws/src/autonomous/src/wall_following.cpp
```diff
@@ -0,0 +1,336 @@
...
+	//std::cout << "correction:" << correction << std::endl;
+	ROS_INFO_STREAM("right correction: " << correction);
+
+	g_right_prev_error = g_right_error;
+
+	// use correction to increment or decrement steering angle
+	g_right_corrected_angle = (correction * M_PI) / 180;
+
+	// boundary check to avoid extrem angle values
+	if(g_right_corrected_angle > 20)
```
Dadie: Why is everything above 20 an extrem angle?

ros_ws/src/autonomous/src/wall_following.cpp
```diff
@@ -0,0 +1,336 @@
...
+	{
+		g_right_corrected_angle = 20;
+	}
+	if(g_right_corrected_angle < -20)
+	{
+		g_right_corrected_angle = -20;
+	}
+
+	float g_right_velocity = 4 - fabs(g_right_corrected_angle);
```
Dadie: Is this the same 4 as the 4 in the `getRange` function?

ros_ws/src/autonomous/src/wall_following.cpp
```diff
@@ -0,0 +1,336 @@
...
+
+	if(g_right_velocity > 4)
```
Dadie: Why is `g_right_velocity` above 4 something that needs to be checked? Can the velocity be -4? If the velocity can't be negative, may switch so `std::min` e.g. `g_right_velocity = std::min(MAGIC_NUMBER_FOUR, g_right_velocity);`

ros_ws/src/autonomous/src/wall_following.cpp
```diff
@@ -0,0 +1,336 @@
...
+	{
+		g_right_velocity = 4;
+	}
+	if(g_right_velocity < 0.3)
```
Dadie: Why is a `g_right_velocity` below 0.3 something that needs to be checked?

ros_ws/src/autonomous/src/wall_following.cpp
```diff
@@ -0,0 +1,336 @@
...
+ * @param lidar 
+ * @return std::array<float, 2> 
+ */
+std::array<float, 2> followLeftWall(const sensor_msgs::LaserScan::ConstPtr& lidar)
+{
+	// calculations done according to the F1Tenth tutorial
+	// http://f1tenth.org/lab_instructions/t6.pdf
+
+	// some angle theta between 0 and 70 which you can choose by yourself
+	float theta = 175;
```
Dadie: Why is the angle theta 175? Also does not the comment say it should be between 0 and 70? What does 0 and 70 mean?

ros_ws/src/teleoperation/include/keyboard_controller.h
```diff
@@ -1,37 +1,87 @@
 #pragma once
```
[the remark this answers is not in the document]
DigitalWayfarer: I don't know, now I see it too. I didn't put them there ^^'

PR #120, opened 2018-12-20
dms-controller

Sami19944:
implemented the dms controller.
keyboard support should be implemented my @marian42 
gamepad support incomming @StefanJVA

### DIFF
--- a/ros_ws/src/car_control/src/car_controller.cpp
+++ b/ros_ws/src/car_control/src/car_controller.cpp
@@ -1,26 +1,19 @@
 #include "car_controller.h"
 #include "car_config.h"
-
-#include "car_control.h"
-
-#include "car_control.h"
-
-#include "car_control.h"
-
 #include "car_control.h"
 
 CarController::CarController()
-    : enabled{ false }
+    : m_enabled{ false }
 {
-    this->drive_parameters_subscriber =
-        this->node_handle.subscribe<drive_msgs::drive_param>(TOPIC_DRIVE_PARAM, 1,
-                                                             &CarController::driveParametersCallback, this);
-    this->command_subscriber =
-        this->node_handle.subscribe<std_msgs::String>(TOPIC_COMMAND, 1, &CarController::commandCallback, this);
+    this->m_drive_parameters_subscriber =
+        this->m_node_handle.subscribe<drive_msgs::drive_param>(TOPIC_DRIVE_PARAM, 1,
+                                                               &CarController::driveParametersCallback, this);
+    this->m_command_subscriber =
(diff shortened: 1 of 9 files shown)


ros_ws/src/car_control/src/car_controller.cpp
```diff
@@ -32,11 +33,7 @@ void CarController::publishDriveParameters(double raw_speed, double raw_angle)
         angle_message.data = angle;
         this->angle_publisher.publish(angle_message);
     }
-    else
-    {
-        std::cout << "not running - ";
-    }
-    std::cout << "speed: " << speed << " | angle: " << angle << std::endl;
+    std::cout << "running: " << this->enabled << " | speed: " << speed << " | angle: " << angle << std::endl;
```
Dadie: Could you replace this call with a [ROS_{DEBUG|INFO|WARNING|...}_STREAM](wiki.ros.org/rosconsole) call?
E.g.
```C++
ROS_DEBUG_STREAM("running: " << this->enabled << " | speed: " << speed << " | angle: " << angle << std::endl);
```

ros_ws/src/car_control/src/dms_controller.cpp
```diff
@@ -0,0 +1,73 @@
...
+
+            m_running = false;
+            std_msgs::String command_message;
+            command_message.data = "stop";
+            this->m_command_pulisher.publish(command_message);
+        }
+        // dms works fine
+        else
+        {
+        }
```
Dadie: What is the purpose of this else?

ros_ws/src/car_control/src/dms_controller.cpp
```diff
@@ -0,0 +1,73 @@
...
+    }
+    if (!m_running)
```
Dadie: is it intended that this codes runs if `m_running` was at the beginning of the function call `true` and was set to `false` in line 27? If yes, please add an comment explaining what the block above (the `if(m_running)` one) does and what this block does. If no, might es well make a simple `else` out of the `if`

ros_ws/src/car_control/src/dms_controller.cpp
```diff
@@ -0,0 +1,73 @@
...
+    {
+        // dms is up-to-date
+        if (m_last_dms_message_received + DMS_EXPIRATION >= timestamp)
+        {
+
+            m_running = true;
+            std_msgs::String command_message;
+            command_message.data = "go";
```
Dadie: Would it be possible to define the commands ("go" and "stop") somewhere in a global header file which is included by the DMS and by the CarController?

PR #121, opened 2018-12-20
Reverse driving and an invert steering option for the gamepad controller

marian42:
- Allows you to drive backwards with the left trigger
- Allows you to configure whether steering is inverted in the launch file, or at runtime using the X button on the gamepad

Before merging, this needs to be tested on our real hardware.

### DIFF
--- a/ros_ws/src/teleoperation/launch/remote_control.launch
+++ b/ros_ws/src/teleoperation/launch/remote_control.launch
@@ -11,6 +11,7 @@ Launches all necessary nodes that let the user control the car with the keyboard
       type="joystick_controller"
       name="joystick_controller"
       output="log" >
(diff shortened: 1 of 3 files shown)


ros_ws/src/teleoperation/launch/remote_control.launch
```diff
@@ -11,6 +11,7 @@ Launches all necessary nodes that let the user control the car with the keyboard
       type="joystick_controller"
       name="joystick_controller"
       output="log" >
+      <param name="invert_steering" type="bool" value ="true" />
```
Nils-Jung: `<param>` is used to hard-code parameters which is probably not what we want here. A better way would be to use `<arg>` here:
```suggestion
       <arg name="invert_steering" default ="true" />
```

PR #123, opened 2018-12-21
Added odometry topic for gazebo simulation

Nils-Jung:
implements #111 

The simulation now publishes a topic called "/odom" which contains all odometry data.
The implementation is based on [this python script from the racecar_gazebo project](https://github.com/mit-racecar/racecar-simulator/blob/master/racecar_gazebo/scripts/gazebo_odometry.py).

### DIFF
--- a/ros_ws/src/simulation/racer_sensors/include/racer_odometry.h
+++ b/ros_ws/src/simulation/racer_sensors/include/racer_odometry.h
@@ -1,0 +1,37 @@
+#pragma once
+
+#include <boost/bind.hpp>
+
+#include <ros/ros.h>
+
+#include <gazebo_msgs/LinkStates.h>
+#include <geometry_msgs/Pose.h>
+#include <geometry_msgs/Twist.h>
+#include <nav_msgs/Odometry.h>
+#include <std_msgs/Header.h>
+
+#define TOPIC_GAZEBO_ODOM "/gazebo/link_states"
+#define TOPIC_ODOM "/odom"
+
+#define BASE_LINK_NAME "racer::base_link"
+
+class RacerOdometry
+{
+    public:
+    RacerOdometry();
+
+    private:
+    ros::NodeHandle m_node_handle;
+
+    ros::Subscriber m_gazebo_subscriber;
+    ros::Publisher m_odom_publisher;
+
+    geometry_msgs::Pose m_last_received_pose;
+    geometry_msgs::Twist m_last_received_twist;
+    ros::Time m_last_received_stamp;
+
+    ros::Timer m_timer;
+
+    void robotPoseUpdate(const gazebo_msgs::LinkStates::ConstPtr& links);
+    void timerCallback(const ros::TimerEvent&);
+};
--- a/ros_ws/src/simulation/racer_sensors/include/racer_odometry.h
+++ b/ros_ws/src/simulation/racer_sensors/include/racer_odometry.h
@@ -10,28 +10,75 @@
 #include <nav_msgs/Odometry.h>
 #include <std_msgs/Header.h>
 
-#define TOPIC_GAZEBO_ODOM "/gazebo/link_states"
-#define TOPIC_ODOM "/odom"
+/**
+ * @brief Rostopic to which Gazebo publishes information about the models of the simulation.
+ */
+constexpr const char* TOPIC_GAZEBO_ODOM = "/gazebo/link_states";
 
-#define BASE_LINK_NAME "racer::base_link"
+/**
+ * @brief Rostopic to which the odometry data should be published to.
+ *
+ */
+constexpr const char* TOPIC_ODOM = "/odom";
+
+/**
+ * @brief Name of the link of the simulated car chassis.
+ *
+ */
+constexpr const char* BASE_LINK_NAME = "racer::base_link";
 
 class RacerOdometry
 {
     public:
+    /**
+     * @brief Construcs a new Racer Odometry object.
+     */
     RacerOdometry();
 
     private:
+    /**
+     * @brief ROS node handle, used for communication with the topics.
+     */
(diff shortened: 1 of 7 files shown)


ros_ws/src/simulation/racer_sensors/src/racer_odometry.cpp
```diff
@@ -0,0 +1,48 @@
...
+        }
+    }
+}
+
+void RacerOdometry::timerCallback(const ros::TimerEvent&)
+{
+    nav_msgs::Odometry msg;
+    msg.header.stamp = m_last_received_stamp;
+    msg.header.frame_id = "map";
+    msg.child_frame_id = "odom";
```
Dadie: May move the strings "map" and "odom" in a global constant

ros_ws/src/simulation/racer_sensors/src/racer_odometry.cpp
```diff
@@ -0,0 +1,48 @@
...
+    msg.pose.pose = m_last_received_pose;
+    msg.twist.twist = m_last_received_twist;
+    m_odom_publisher.publish(msg);
```
Dadie: May use scoping to emphasises that the message is filled with data e.g.
```C++
nav_msgs::Odometry msg;
{
    nav_msgs::Odometry msg;
    msg.header.stamp = m_last_received_stamp;
    msg.header.frame_id = "map";
    msg.child_frame_id = "odom";
    msg.pose.pose = m_last_received_pose;
    msg.twist.twist = m_last_received_twist;
}
m_odom_publisher.publish(msg);
```

ros_ws/src/simulation/racer_sensors/src/racer_odometry.cpp
```diff
@@ -0,0 +1,48 @@
...
+        {
+            m_last_received_pose = links->pose.at(i);
+            m_last_received_twist = links->twist.at(i);
+            m_last_received_stamp = ros::Time::now();
+            return;
+        }
+    }
+}
+
+void RacerOdometry::timerCallback(const ros::TimerEvent&)
```
Dadie: Is it possible, that `timerCallback` is called while `robotPoseUpdate` is called in another thread?

ros_ws/src/simulation/racer_sensors/src/racer_odometry.cpp
```diff
@@ -0,0 +1,48 @@
...
+    m_last_received_pose = geometry_msgs::Pose();
+    m_last_received_twist = geometry_msgs::Twist();
+    m_last_received_stamp = ros::Time();
+
+    m_timer = m_node_handle.createTimer(ros::Duration(0.05), boost::bind(&RacerOdometry::timerCallback, this, _1));
+}
+
+void RacerOdometry::robotPoseUpdate(const gazebo_msgs::LinkStates::ConstPtr& links)
+{
+    for (uint i = 0; i < links->name.size(); i++)
```
Dadie: Where does the type `uint` comes from? How is it defined?

ros_ws/src/simulation/racer_sensors/src/racer_odometry.cpp
```diff
@@ -0,0 +1,48 @@
...
+    nav_msgs::Odometry msg;
+    msg.header.stamp = m_last_received_stamp;
+    msg.header.frame_id = "map";
+    msg.child_frame_id = "odom";
+    msg.pose.pose = m_last_received_pose;
+    msg.twist.twist = m_last_received_twist;
+    m_odom_publisher.publish(msg);
+}
+
+int main(int argc, char** argv)
```
Dadie: May split the classes implementation from the program.

ros_ws/src/simulation/racer_sensors/include/racer_odometry.h
```diff
@@ -0,0 +1,37 @@
...
+#include <ros/ros.h>
+
+#include <gazebo_msgs/LinkStates.h>
+#include <geometry_msgs/Pose.h>
+#include <geometry_msgs/Twist.h>
+#include <nav_msgs/Odometry.h>
+#include <std_msgs/Header.h>
+
+#define TOPIC_GAZEBO_ODOM "/gazebo/link_states"
+#define TOPIC_ODOM "/odom"
```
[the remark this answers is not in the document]
Nils-Jung: Done. I'll change the wiki once this is on the master branch.

ros_ws/src/simulation/racer_sensors/src/racer_odometry.cpp
Nils-Jung: Removed those strings since they don't seem to do anything. We can put them back if we ever need them.

ros_ws/src/simulation/racer_sensors/src/racer_odometry.cpp
Nils-Jung: Done.

ros_ws/src/simulation/racer_sensors/src/racer_odometry.cpp
Nils-Jung: No. `ROS:spin()` only works on a single thread, so that is not possible (it is possible to do this multithreaded but we don't need it here).

ros_ws/src/simulation/racer_sensors/src/racer_odometry.cpp
Nils-Jung: Changed `uint` to `unsigned int`. They are synonymous but I get that it is bad form.

ros_ws/src/simulation/racer_sensors/src/racer_odometry.cpp
Nils-Jung: Done. The main function now resides in `main_racer_odometry.cpp`.

PR #137, opened 2019-01-15
joystick mapping for xbox360 and ps3 gamepads

StefanJVA:
Somehow the ps3 mappings on my pc seem to be different from the ones on the actual car.
Therefore we have to make the ps3 mappings on Friday. So don't merge this yet!
However the code can be reviewed any time.


ros_ws/src/teleoperation/src/joystick_controller.cpp
```diff
@@ -14,8 +17,21 @@ JoystickController::JoystickController()
...
 
+
+    // get the provided gamepad type
     ros::NodeHandle private_node_handle("~");
-    private_node_handle.getParam(INVERT_STEERING_PARAMETER, this->m_invert_steering);
+    private_node_handle.getParam(GAMEPAD_TYPE_PARAMETER, this->m_gamepad_type);
+
+    // load joystick map for the provided gamepad type
+    ROS_ASSERT_MSG(m_gamepad_type != "xbox360" || m_gamepad_type != "ps3", "Gamepad type should be xbox360 or ps3");
+    if (m_gamepad_type == "xbox360")
```
marian42: Is this how string comparisons are done in C++? Afaik, this shouldn't work (use something like strcmp instead). Did you test your code with both controllers?

ros_ws/src/teleoperation/include/joystick_map.h
```diff
@@ -0,0 +1,47 @@
...
+    virtual ~JoystickMap()
+    {
+    }
+
+    /**
+     * @brief Get value of the steering axis. -1 is maximal left, 1 is maximal right
+     *
+     * @return float
+     */
+    virtual float getSteeringAxis(const sensor_msgs::Joy::ConstPtr& joystick) = 0;
```
marian42: The method name and summary don't match. Does this return the steering axis or its value? Does axis refer to the car or the joystick? It should probably be named `getSteering`.

ros_ws/src/teleoperation/include/joystick_map.h
```diff
@@ -0,0 +1,47 @@
...
+{
+    protected:
+    JoystickMap()
+    {
+    }
+
+    public:
+    virtual ~JoystickMap()
+    {
+    }
```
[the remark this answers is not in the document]
Dadie: A virtual class should always have a virtual destructor. It can be omitted, but should not. Also instead of
```C++
virtual ~JoystickMap()
{
}
```
one could write
```C++
virtual ~JoystickMap() = default;
```

ros_ws/src/teleoperation/include/joystick_map_ps3.h
```diff
@@ -0,0 +1,33 @@
...
+class JoystickMapPs3 : public JoystickMap
+{
+    public:
+    JoystickMapPs3()
+    {
+    }
+
+    ~JoystickMapPs3()
+    {
+    }
```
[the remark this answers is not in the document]
Dadie: As the virtual base class has already a destructor this one can be omitted AFAIK.

ros_ws/src/teleoperation/include/joystick_map_ps3.h
```diff
@@ -0,0 +1,33 @@
+#include "joystick_map.h"
+
+class JoystickMapPs3 : public JoystickMap
+{
+    public:
+    JoystickMapPs3()
```
Dadie: If the constructor does really nothing, you could write
```C++
JoystickMapPs3() = default;
```

ros_ws/src/teleoperation/include/joystick_map_ps3.h
```diff
@@ -0,0 +1,33 @@
...
+    {
+    }
+
+    ~JoystickMapPs3()
+    {
+    }
+
+    float getSteeringAxis(const sensor_msgs::Joy::ConstPtr& joystick)
```
Dadie: **MAY**
Add the virtual keyword to make it more clear, that this is a virtual member function.

ros_ws/src/teleoperation/src/joystick_controller.cpp
Dadie: This type of comparison works most of the time but yes it shouldn't work. Thats why we got the (std::string literal)[https://en.cppreference.com/w/cpp/string/basic_string/operator""s] in C++11.

ros_ws/src/teleoperation/include/joystick_map_ps3.h
```diff
@@ -0,0 +1,33 @@
...
+    {
```
Dadie: **MAY**
If you implement a member function in a header file you may want to add the `inline` keyword as otherwise the same member function might be implemented twice by different source files (of both include this header file) which leads to link time errors.

ros_ws/src/teleoperation/src/joystick_controller.cpp
```diff
@@ -14,8 +17,21 @@ JoystickController::JoystickController()
...
+    {
+        m_joystick_map = new JoystickMapXbox360();
```
Dadie: With the `new` you create a memory leak. Either add a `delete m_joystick_map;` to the destructor or use ownership-semantics by changing the type of  `m_joystick_map` to `std::unique_ptr<JoystickController>` and replace the `new JoystickMapXbox360();` with `std::make_unique<JoystickMapXbox360>();`

ros_ws/src/teleoperation/src/joystick_controller.cpp
```diff
@@ -14,8 +17,21 @@ JoystickController::JoystickController()
...
+    }
+    else if (m_gamepad_type == "ps3")
+    {
+        m_joystick_map = new JoystickMapPs3();
+    }
```
Dadie: **MAY**
Add an `else` case that throws an exception, outputs an error or gives any kind of indication that something happened that should not have happened.

ros_ws/src/teleoperation/src/joystick_controller.cpp
```diff
@@ -14,8 +17,21 @@ JoystickController::JoystickController()
     this->m_joystick_subscriber =
         this->m_node_handle.subscribe<sensor_msgs::Joy>("joy", 10, &JoystickController::joystickCallback, this);
 
+
+    // get the provided gamepad type
```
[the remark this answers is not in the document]
Dadie: I think the comment can be quite useful as is puts the next few lines of code under the umbrella of "get the provided gamepad type". And as someone that hasn't worked quite that much with ROS I appreciated the comment as I find the whole `getParam` stuff quite confusing.

ros_ws/src/teleoperation/src/joystick_controller.cpp
```diff
@@ -40,19 +55,12 @@ void JoystickController::joystickCallback(const sensor_msgs::Joy::ConstPtr& joys
...
+    float steering_angle = m_joystick_map->getSteeringAxis(joystick) * STEERING_SCALING_FACTOR;
 
-    // Detect if the button was pressed since the last reading
-    bool invert_toggle_button = joystick->buttons[JOYSTICK_BUTTON_TOGGLE_INVERT_STEERING] == 1;
-    if (invert_toggle_button && !this->m_toggle_invert_steering_state)
-    {
-        this->m_invert_steering = !this->m_invert_steering;
-    }
-    this->m_toggle_invert_steering_state = invert_toggle_button;
+    this->publishDriveParameters(acceleration - deceleration, steering_angle);
```
Dadie: **MAY**
Before publishing the parameters a final sanity check might be useful like are the values for the `acceleration` in the expected range, is the difference between `acceleration` and `deceleration` in the expected range.

ros_ws/src/teleoperation/src/joystick_controller.cpp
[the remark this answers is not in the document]
marian42: The comment only refers to the next line (not the next few lines). All the words in the comment are also in the next line of code (except "the" and "provided"), so there is literally no additiona information. Comments should not say what is being done, but why it is being done.

ros_ws/src/teleoperation/include/joystick_map.h
```diff
@@ -0,0 +1,47 @@
+#pragma once
+
+#include <ros/ros.h>
+
+#include <sensor_msgs/Joy.h>
+
+class JoystickMap
```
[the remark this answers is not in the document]
Dadie: I agree with marian here and also think `AbstractJoystickMap` would be a good name.

marian42:
Now that we have set the correct values for three different controllers, it is apparent that only the axis numbers change, while everything else stays the same. I think the best approach would be to store the axis assignments like this:

```
struct JoystickMapping
{
    int steeringAxis;
    int AccelerationAxis;
    int DecelerationAxis;
    int DeadMansSwitchButton;
}
```

You could then define three constant values of that struct and select one at runtime.

StefanJVA:
> Now that we have set the correct values for three different controllers, it is apparent that only the axis numbers change, while everything else stays the same. I think the best approach would be to store the axis assignments like this:
> 
> ```
> struct JoystickMapping
> {
>     int steeringAxis;
>     int accelerationAxis;
>     int decelerationAxis;
>     int deadMansSwitchButton;
> }
> ```
> 
> You could then define three constant values of that struct and select one at runtime.

I have spend 30 minutes fixing every suggestion that was made, only to arrive at this comment, which makes everything obsolete anyway. FML :(

ros_ws/src/teleoperation/src/joystick_controller.cpp
StefanJVA: Actually i am still not sure how to compare strings properly in c++

ros_ws/src/teleoperation/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ cmake_minimum_required(VERSION 2.8.3)
 project(teleoperation)
 
 ## Compile as C++11, supported in ROS Kinetic and newer
-set (CMAKE_CXX_STANDARD 11)
+add_compile_options(-std=c++11)
```
Dadie: Why did you change this? Setting the `CMAKE_CXX_STANDARD` should be prefered over setting the Argument manually.

ros_ws/src/teleoperation/src/joystick_controller.cpp
Dadie: Sorry, my fault. == is fine ^^

ros_ws/src/teleoperation/CMakeLists.txt
StefanJVA: because std::unique_ptr

PR #150, opened 2019-01-18
dms-controller - changed some constants to parameters

Sami19944:


ros_ws/src/car_control/src/dms_controller.cpp
```diff
@@ -11,6 +11,9 @@ DMSController::DMSController()
         this->m_node_handle.subscribe<std_msgs::Int64>(TOPIC_DMS, 1, &DMSController::dmsCallback, this);
 
     this->m_command_pulisher = this->m_node_handle.advertise<std_msgs::String>(TOPIC_COMMAND, 1);
+
+    m_node_handle.getParam(DMS_CHECK_RATE_PARAMETER, this->dms_check_rate);
+    m_node_handle.getParam(DMS_EXPIRATION_PARAMETER, this->dms_expiration);
```
Dadie: **MAY**
Add a sanity check e.g.
- Is any value zero or below?
- In any value way to high?

ros_ws/src/car_control/include/dms_controller.h
```diff
@@ -5,17 +5,16 @@
 #include <std_msgs/Int64.h>
 #include <std_msgs/String.h>
 
-// How often the dead mans switch is checked. in Hz
-constexpr const int DMS_CHECK_RATE = 20;
-
-// How old the last dead mans switch check can be. in ms
-constexpr const int DMS_EXPIRATION = 100;
+constexpr const char* DMS_CHECK_RATE_PARAMETER = "/dms_controller/dms_check_rate";
```
Dadie: **MAY**
Add the Prefix "TOPIC_" to emphasis both are topics.

Sami19944: These are not topics, but parameter names.

Dadie: In that case it is kind of weird that we (as far as I can see in the code base) use "TOPIC_" as a prefix for a topic **but** "_PARAMETER" as a suffix for a parameter.

ros_ws/src/car_control/src/dms_controller.cpp
```diff
@@ -11,7 +11,19 @@ DMSController::DMSController()
         this->m_node_handle.subscribe<std_msgs::Int64>(TOPIC_DMS, 1, &DMSController::dmsCallback, this);
 
     this->m_command_pulisher = this->m_node_handle.advertise<std_msgs::String>(TOPIC_COMMAND, 1);
-}
+    
+    m_node_handle.getParam(PARAMETER_DMS_CHECK_RATE, this->dms_check_rate);
```
StefanJVA:
```suggestion
    ros::NodeHandle private_node_handle("~");
    private_node_handle.getParam(PARAMETER_DMS_CHECK_RATE, this->dms_check_rate);
```

ros_ws/src/car_control/src/dms_controller.cpp
```diff
@@ -11,7 +11,19 @@ DMSController::DMSController()
...
+	if(dms_check_rate <= 0 || dms_check_rate > 1000) {
+		ROS_WARN_STREAM("dms_check_rate should be bigger than 0. dms_check_rate is now set to 20 (hz)");
+		dms_check_rate = 20;
+	}
+	
+    m_node_handle.getParam(PARAMETER_DMS_EXPIRATION, this->dms_expiration);
```
StefanJVA:
```suggestion
    private_node_handle.getParam(PARAMETER_DMS_EXPIRATION, this->dms_expiration);
```

Sami19944: It works, I tested it

ros_ws/src/car_control/src/dms_controller.cpp
Sami19944: The nodehandler is a member variable. I tested it and it works fine.

ros_ws/src/car_control/include/dms_controller.h
```diff
@@ -6,15 +6,17 @@
 #include <std_msgs/Int64.h>
 #include <std_msgs/String.h>
 
-// How often the dead mans switch is checked. in Hz
-constexpr const int DMS_CHECK_RATE = 20;
+constexpr const char* PARAMETER_DMS_CHECK_RATE = "dms_check_rate"; //20
+constexpr const char* PARAMETER_DMS_EXPIRATION = "dms_expiration"; //100
```
marian42: These comments are very confusing. Please remove them or write something like "Default value is 20Hz". Also doesn't writing the value into the code as a comment defeat the purpose of having them as parameters? If the value is in the code, you need to change it whenever changing the config file.

Sami19944: The DMS_CHECK_RATE constant isnt in the code.

ros_ws/src/car_control/launch/car_control.launch
```diff
@@ -20,6 +20,8 @@ Launches all necessary nodes that listen for drive_parameters and translate them
       type="dms_controller"
       name="dms_controller"
       output="screen" >
+      <param name="dms_check_rate" type="int" value = "20" /> <!-- in Hz -->
+      <param name="dms_expiration" type="int" value = "100" /> <!-- in ms -->
```
Nils-Jung: Maybe add <arg> tags so these parameters can be changed from the command line?

StefanJVA: > Do you have an example?

Look into src/teleoperation/launch/remote_control.launch

Sami19944:
sry, it works now

PR #163, opened 2019-02-01
Alex cmd_vel

FlyingdutchmanC:


ros_ws/src/navigation_stack/msgs_converter/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+## Specify additional locations of header files
+include_directories(
+  include
+  ${catkin_INCLUDE_DIRS}
+)
+
+# Declare joystick controller node executable
+add_executable(msgs_converter
+src/main.cpp
+src/msgs_converter.cpp
```
marian42: These could be indented for consistency.

ros_ws/src/navigation_stack/msgs_converter/include/msgs_converter.h
```diff
@@ -0,0 +1,33 @@
...
+class MSGSConverter
+{
+    public:
+    MSGSConverter();
+
+    private:
+    /**
+     * @brief ROS Handle
+     *
+     */
```
marian42: This summary doesn't contain any information. It should be removed or expanded.

ros_ws/src/navigation_stack/msgs_converter/include/msgs_converter.h
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+#include <ros/ros.h>
+
+#include <drive_msgs/drive_param.h>
+#include <geometry_msgs/Twist.h>
+
+class MSGSConverter
```
marian42: Can this be renamed to avoid the acronym? Either way, a summary would be helpful here to explain what the class does and why it's needed.

ros_ws/src/navigation_stack/msgs_converter/include/msgs_converter.h
```diff
@@ -0,0 +1,33 @@
...
+    ros::NodeHandle m_node_handle;
+
+    /**
+     * @brief ROS Subscriber for servo position messages
+     * TOPIC: "cmd_vel"
+     */
+    ros::Subscriber m_command_velocity_subscriber;
+
+    /**
+     * @brief ROS Publischer for the velocity of left rear wheel
```
marian42:
```suggestion
     * @brief ROS publisher for the velocity of left rear wheel
```

ros_ws/src/navigation_stack/msgs_converter/include/msgs_converter.h
```diff
@@ -0,0 +1,33 @@
...
+     *
+     */
+    ros::Publisher m_to_car_control_publisher;
```
marian42: Could this be renamed to something like `m_left_rear_wheel_velocity_publisher`? I didn't quite understand what this does. Maybe you can update the summary to explain why only the left rear wheel velocity needs to be published.

ros_ws/src/navigation_stack/msgs_converter/src/msgs_converter.cpp
```diff
@@ -0,0 +1,60 @@
...
+{
+
+    double velocity_x = cmd_vel_message->linear.x;
+    double velocity_y = cmd_vel_message->linear.y;
+    double velocity_angular = cmd_vel_message->angular.z;
+
+    Vector2d metric_velocity(velocity_x, velocity_y);
+
+    double velocity_result =
+        metric_velocity.norm(); // check if norm is zero, then the steering angle (angle_rad) will be infinity!
```
marian42: The comment is confusing as this line does not check if the velocity is zero.

ros_ws/src/navigation_stack/msgs_converter/src/msgs_converter.cpp
```diff
@@ -0,0 +1,60 @@
...
+    double erpm_speed = metric_velocity.norm() * car_config::TRANSMISSION / car_config::ERPM_TO_SPEED;
```
marian42: Instead of `metric_velocity.norm()` you could use the variable from the line above.

ros_ws/src/navigation_stack/msgs_converter/CMakeLists.txt
```diff
@@ -0,0 +1,48 @@
...
+  drive_msgs
+  roscpp
+  rospy
+  std_msgs
+  vesc_msgs
+  nav_msgs
+  vesc_sim
+)
+
+
```
marian42: Please add the line that sets c++ warnings from one of the other CMake files, such as `ros_ws/src/navigation_stack/msgs_converter/CMakeLists.txt`.

ros_ws/src/navigation_stack/msgs_converter/src/msgs_converter.cpp
```diff
@@ -0,0 +1,60 @@
...
+    double angle_rad = 0;
+
+    if (velocity_angular != 0)
```
marian42: With the updated cmake file, this creates a warning. Please decide wether it's justified and either remove the comparison or suppress the warning. I think in this case it would be best to do something like `abs(velocity_angluar) < SMALL_NUMBER`.

ros_ws/src/navigation_stack/msgs_converter/src/msgs_converter.cpp
```diff
@@ -0,0 +1,60 @@
...
+    {
+        if (velocity_result < 0 && velocity_result > -0.001)
```
marian42: I think it would make sense to define a constant like `constexpr VELOCITY_THRESHOLD = 0.001`.

ros_ws/src/navigation_stack/msgs_converter/src/msgs_converter.cpp
```diff
@@ -0,0 +1,60 @@
...
+        {
+            velocity_result = -0.001;
+        }
+        else if (velocity_result > 0 && velocity_result < 0.001)
+        {
+            velocity_result = -0.001;
```
marian42: Shouldn't the velocity be set to something positive here?

ros_ws/src/navigation_stack/msgs_converter/src/msgs_converter.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+    double servo_data = (angle_rad * car_config::STEERING_TO_SERVO_GAIN) + car_config::STEERING_TO_SERVO_OFFSET;
+    if (servo_data < 0)
+    {
+        servo_data = 0;
+    }
+    else if (servo_data > 1)
+    {
+        servo_data = 1;
+    }
```
marian42: You can replace this with `servo_data = std::clamp(servo_data, 0, 1)`.

ros_ws/src/navigation_stack/msgs_converter/src/msgs_converter.cpp
```diff
@@ -0,0 +1,60 @@
...
+
+    drive_msgs::drive_param control;
```
marian42: This could be renamed to `control_message`.

PR #174, opened 2019-03-01
velocity and angle check

Sami19944:
see code changes (small)

### DIFF
--- a/ros_ws/src/car_control/src/car_controller.cpp
+++ b/ros_ws/src/car_control/src/car_controller.cpp
@@ -2,6 +2,8 @@
 #include "car_config.h"
 #include "car_control.h"
 
+#include <boost/algorithm/clamp.hpp>
+
 CarController::CarController()
     : m_enabled{ false }
 {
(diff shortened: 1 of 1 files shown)


marian42:
You implemented a clamp function. C++ has a built-in clamp function since C++17, but we can't use that because we're on an older version. We have clamp functions in our project [here](https://github.com/Autonomous-Racing-PG/ros.package/blob/master/ros_ws/src/teleoperation/src/keyboard_controller.cpp#L5) and [here](https://github.com/Autonomous-Racing-PG/ros.package/blob/master/ros_ws/src/navigation_stack/msgs_converter/src/navigation_stack_control_converter.cpp#L53). I'm not sure what is a good way to go about this. Can/Should we define a clamp function in one place and then reuse it? It seems like boost has a clamp function and our project already depends on boost so we could use that. @Dadie what do you think?

Dadie:
As we already have an dependency on boost (by ROS), [`boost::algorithm::clamp`](https://www.boost.org/doc/libs/1_67_0/libs/algorithm/doc/html/header/boost/algorithm/clamp_hpp.html) could be used.

E.g.
```C++
#include <boost/algorithm/clamp.hpp>

void foo() {
    // DO STUFF [..]
    auto value = 100;
    auto low = std::numeric_limits<typeof(value)>::min();
    auto high = std::numeric_limits<typeof(value)>::max();
    auto clamped_value = boost::algorithm::clamp(value, low, high);
    // DO MORE STUFF [..]
}
```

PR #182, opened 2019-03-06
Refactor Wallfollowing

marian42:
This pullrequest tries to make the wallfollowing code more readable and more maintainable.
The updated code behaves like the previous wallfollowing node, actual improvements will come in a later pullrequest.


ros_ws/src/autonomous/src/wall_following.cpp
```diff
@@ -1,202 +1,90 @@
...
-    float range = lidar->ranges[theta / 0.375];
-
-    // check for absurd or nan values
-    // the max range should be 4 so the car doesn't get affected by walls very far away
-    if (range > 4 || std::isinf(range))
-    {
-        range = 4;
-    }
-    return range;
+    return out_lower + (out_upper - out_lower) * (value - in_lower) / (in_upper - in_lower);
```
ce-sch: Not necessary but makes it clearer on first glance:
```suggestion
    return out_lower + ((out_upper - out_lower) * (value - in_lower) / (in_upper - in_lower));
```

ros_ws/src/autonomous/src/wall_following.cpp
```diff
@@ -1,202 +1,90 @@
...
-    emer_stop_subscriber =
-        m_node_handle.subscribe<std_msgs::Bool>(TOPIC_EMER_STOP, 1, &WallFollowing::emergencyStopCallback, this);
-    pid_publisher = m_node_handle.advertise<drive_msgs::pid_input>(TOPIC_PID_INPUT, 1);
+    this->m_emergency_stop_subscriber =
+        m_node_handle.subscribe<std_msgs::Bool>(TOPIC_EMERGENCY_STOP, 1, &WallFollowing::emergencyStopCallback, this);
+    this->m_drive_parameter_publisher = m_node_handle.advertise<drive_msgs::drive_param>(TOPIC_DRIVE_PARAMETERS, 1);
 }
 
-float WallFollowing::rangeAtDegree(const sensor_msgs::LaserScan::ConstPtr& lidar, float theta)
+float map(float in_lower, float in_upper, float out_lower, float out_upper, float value)
```
ce-sch:
```suggestion
// map value in range [in_lower, in_upper] to the corresponding number in range [out_lower, out_upper] 
float map(float in_lower, float in_upper, float out_lower, float out_upper, float value)
```

ros_ws/src/autonomous/include/wall_following.h
```diff
@@ -4,95 +4,68 @@
...
+constexpr const char* TOPIC_LASER_SCAN = "/scan";
+constexpr const char* TOPIC_EMERGENCY_STOP = "/emergency_stop";
+
+constexpr int LIDAR_SAMPLE_COUNT = 720;
+
+// Discard lidar measurements outside this range
+constexpr float MIN_RANGE = 0.2;
+constexpr float MAX_RANGE = 30;
+
+constexpr float FALLBACK_RANGE = 4;
```
ce-sch: This is not fail-safe. If the range a undefined lidar-index is requested, the car should not drive on without noticing that something went wrong. It would be safer to set this value to 0.
See also emergency_stop.cpp line 16: rangeAtDegree(). Maybe there should also be used this const?
```suggestion
constexpr float FALLBACK_RANGE = 0;
```

marian42: I agree that the fallback value is problematic. Setting it to zero is not good either, since the point of the min and max values it that you can expect the number to be between them. 0 will lead to a division by zero in the wallfollowing code.

Anyway, the point of this pullrequest is to make more readable what happens in the wallfollowing code without changing its behaviour. I will soon make a second pulllrequest where fallbacks aren't used.

PR #183, opened 2019-03-07
Wallfollowing improvements

marian42:
- Visualize detected wall in rviz
- Don't use a hardcoded number for the amount of samples in the lidar since it's different on the simulated and real car
- Implement double wall following

Currently, code changes are required to switch between following the left, right or both walls. This should be changed once we know how these modes perform.

### DIFF
--- a/ros_ws/src/autonomous/src/wall.cpp
+++ b/ros_ws/src/autonomous/src/wall.cpp
@@ -1,0 +1,20 @@
+#include "wall.h"
+
+float Wall::getAngle()
+{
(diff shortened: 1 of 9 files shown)


ros_ws/src/autonomous/src/wall.cpp
```diff
@@ -0,0 +1,20 @@
+#include "wall.h"
+
+float Wall::getAngle()
+{
+    return atan((m_range1 * cos(std::abs(m_angle1 - m_angle2)) - m_range2) / m_range1 *
```
StefanJVA: Make sure that the used variables are initialized, by (for example) adding an ASSERT or initializing them in wall.h
Same thing goes for the two other methods of this class.

PR #187, opened 2019-03-11
Small C++ cleanup

Dadie:
I skimmed over the whole code base and cleanup (very) small things that catch my eye.

### DIFF
--- a/ros_ws/src/autonomous/src/emergency_stop.cpp
+++ b/ros_ws/src/autonomous/src/emergency_stop.cpp
@@ -51,22 +51,18 @@ bool EmergencyStop::emergencyStop(const sensor_msgs::LaserScan::ConstPtr& lidar)
 
 void EmergencyStop::lidarCallback(const sensor_msgs::LaserScan::ConstPtr& lidar)
 {
-
     bool emergency_stop = emergencyStop(lidar);
 
-    if (emergency_stop == false)
-    {
-        std_msgs::Bool emer_stop;
-        emer_stop.data = false;
-        emer_stop_publisher.publish(emer_stop);
-    }
-    else
+    if (emergency_stop)
     {
         ROS_INFO_STREAM("Too close! Stop!");
-        std_msgs::Bool emer_stop;
-        emer_stop.data = true;
-        emer_stop_publisher.publish(emer_stop);
     }
+
+    std_msgs::Bool emer_stop;
+    {
+        emer_stop.data = emergency_stop;
+    }
+    emer_stop_publisher.publish(emer_stop);
 }
 
 int main(int argc, char** argv)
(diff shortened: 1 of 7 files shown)


ros_ws/src/autonomous/src/emergency_stop.cpp
```diff
@@ -9,23 +9,34 @@ EmergencyStop::EmergencyStop()
 
 float EmergencyStop::rangeAtDegree(const sensor_msgs::LaserScan::ConstPtr& lidar, float theta)
```
marian42: The `rangeAtDegree` function exists twice in our codebase. I made some changes to it, but unfortunately, I only changed the one in `wall_following.cpp`. I think it would be best to reuse that one, but I'm not sure where we should put it.

ros_ws/src/autonomous/src/emergency_stop.cpp
```diff
@@ -46,34 +57,24 @@ bool EmergencyStop::emergencyStop(const sensor_msgs::LaserScan::ConstPtr& lidar)
...
-
     bool emergency_stop = emergencyStop(lidar);
 
-    if (emergency_stop == false)
+    if (emergency_stop)
     {
-        std_msgs::Bool emer_stop;
-        emer_stop.data = false;
-        emer_stop_publisher.publish(emer_stop);
+        // Announce emergency stop
```
marian42: This is not a necesary comment IMO.

ros_ws/src/navigation_stack/msgs_converter/src/navigation_stack_control_converter.cpp
```diff
@@ -48,14 +49,7 @@ void NavigationStackControlConverter::convertCallback(const geometry_msgs::Twist
     double servo_data = (angle_rad * car_config::STEERING_TO_SERVO_GAIN) + car_config::STEERING_TO_SERVO_OFFSET;
 
     // c++17 servo_data = std::clamp(servo_data,0,1);
```
marian42: You should also remove this comment.

ros_ws/src/autonomous/src/emergency_stop.cpp
Dadie: I'm neither sure where to put such a RangeAtDegree function. For now I made the issue #194 and removed the commits changing anything in the `rangeAtDegree` function.

ros_ws/src/autonomous/src/emergency_stop.cpp
StefanJVA: This time i actually agree :D

ros_ws/src/navigation_stack/msgs_converter/src/navigation_stack_control_converter.cpp
Dadie: Comment was removed :smile:

ros_ws/src/autonomous/src/emergency_stop.cpp
Dadie: And... its gone ;)

Dadie:
Rebased on master

Issue #189, opened 2019-03-12
Add Lidar frame parameter to wallfollowing


marian42:
The name of the lidar frame inside the `/scan` message is `hokuyo` in the simulation and `laser` on the real car.
This should become a parameter of the wallfollowing node and be set in the launch files. It's currently hardcoded in `wall_following.cpp` in line 5.

PR #195, opened 2019-03-12
Fixed issue #189

Dadie:
This PR should fix issue #189.

I'm not quite happy with the parameters name, has anyone a better suggestion?

### DIFF
--- a/ros_ws/src/autonomous/include/wall_following.h
+++ b/ros_ws/src/autonomous/include/wall_following.h
@@ -15,6 +15,8 @@
 #include <ros/console.h>
 #include <ros/ros.h>
 
+constexpr const char* PARAMETER_MARKER_FRAME_ID = "marker_frame_id";
+
 constexpr const char* TOPIC_DRIVE_PARAMETERS = "/input/drive_param/wallfollowing";
 constexpr const char* TOPIC_LASER_SCAN = "/scan";
 constexpr const char* TOPIC_EMERGENCY_STOP = "/emergency_stop";
@@ -45,7 +47,7 @@ constexpr float TIME_BETWEEN_SCANS = 0.025;
 class WallFollowing
 {
     public:
-    WallFollowing();
+    WallFollowing(const std::string& frame_id);
 
     private:
     ros::NodeHandle m_node_handle;
--- a/ros_ws/src/autonomous/launch/autonomous_driving.launch
+++ b/ros_ws/src/autonomous/launch/autonomous_driving.launch
@@ -1,6 +1,7 @@
 <?xml version="1.0"?>
 <launch>
 <!-- Launches all necessary nodes that let the car drive autonomously. -->
+    <arg name="marker_frame_id" default="hokuyo"/> <!-- "hokuyo" in the simulation and "laser" on the physical car -->
 
     <!-- wall_following node -->
     <node
@@ -9,6 +10,7 @@
       type="wall_following"
       name="wall_following"
       output="screen" >
+      <param name="marker_frame_id" type="str" value = "$(arg marker_frame_id)" />
     </node>
 
     <!-- emergency_stop node -->
--- a/ros_ws/src/autonomous/src/wall_following.cpp
+++ b/ros_ws/src/autonomous/src/wall_following.cpp
@@ -1,8 +1,8 @@
 #include "wall_following.h"
 #include <boost/algorithm/clamp.hpp>
 
-WallFollowing::WallFollowing()
-    : m_debug_geometry(this->m_node_handle, TOPIC_VISUALIZATION, "hokuyo")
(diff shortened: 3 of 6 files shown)


ros_ws/src/autonomous/src/wall_following.cpp
```diff
@@ -136,7 +137,18 @@ void WallFollowing::lidarCallback(const sensor_msgs::LaserScan::ConstPtr& lidar)
...
+    std::string frame_id;
+    private_node_handle.getParam(PARAMETER_MARKER_FRAME_ID, frame_id);
+    if (frame_id != MARKER_FRAME_ID_SIMULATION_VALUE && frame_id != MARKER_FRAME_ID_CAR_VALUE)
+    {
+        ROS_WARN_STREAM(PARAMETER_MARKER_FRAME_ID << " should be '" << MARKER_FRAME_ID_SIMULATION_VALUE << "' or '"
+                                                  << MARKER_FRAME_ID_CAR_VALUE << "'. Your value: '" << frame_id
+                                                  << "', using '" << MARKER_FRAME_ID_SIMULATION_VALUE << "'.");
+        frame_id = MARKER_FRAME_ID_SIMULATION_VALUE;
+    }
+    WallFollowing wall_following(frame_id);
```
marian42: In our other nodes, the parameters are read in within the node class, not in the main function. I'll leave it to you to decide if it should be changed.
If the parameters are read inside the class, I think a default constructor for the `RvizGeometryPublisher` is needed.

ros_ws/src/autonomous/launch/autonomous_driving.launch
```diff
@@ -1,6 +1,7 @@
 <?xml version="1.0"?>
 <launch>
 <!-- Launches all necessary nodes that let the car drive autonomously. -->
+    <arg name="marker_frame_id" default="hokuyo"/> <!-- "hokuyo" in the simulation and "laser" on Autonomousprime //-->
```
marian42:
```suggestion
    <arg name="marker_frame_id" default="hokuyo"/> <!-- "hokuyo" in the simulation and "laser" on the physical car -->
```
👮 No fun allowed 👮

ros_ws/src/autonomous/include/wall_following.h
```diff
@@ -43,7 +49,7 @@ constexpr float TIME_BETWEEN_SCANS = 0.025;
 class WallFollowing
 {
     public:
-    WallFollowing();
+    WallFollowing(const std::string&);
```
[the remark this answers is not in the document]
Dadie: Fixed.

ros_ws/src/autonomous/include/wall_following.h
```diff
@@ -13,11 +13,17 @@
...
+constexpr const char* PARAMETER_MARKER_FRAME_ID = "marker_frame_id";
+
 constexpr const char* TOPIC_DRIVE_PARAMETERS = "/input/drive_param/wallfollowing";
 constexpr const char* TOPIC_LASER_SCAN = "/scan";
 constexpr const char* TOPIC_EMERGENCY_STOP = "/emergency_stop";
 constexpr const char* TOPIC_VISUALIZATION = "/wallfollowing_visualization";
 
+// The parameter PARAMETER_MARKER_FRAME_ID should be either the SIMULATION or (real) CAR VALUE.
+constexpr const char* MARKER_FRAME_ID_SIMULATION_VALUE = "hokuyo";
+constexpr const char* MARKER_FRAME_ID_CAR_VALUE = "laser";
```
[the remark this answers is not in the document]
Dadie: You're right. I still added at least an empty string check because empty string should be in almost any case wrong.

ros_ws/src/autonomous/src/wall_following.cpp
Dadie: At first I also thought about putting it in the constructor, but for two reasons I did not do so.

1. There are (IMO) no always valid default values for the arguments given to the current constructor of the `RvizGeometryPublisher` class. So to whatever a default constructor would set the members, it would be in almost any case the wrong values. By not having a default constructor a `RvizGeometryPublisher` object can't be created by accident with invalid values, but only by actively give it the arguments to the constructor thus make it more robust, easier to debug and (IMO) less prone for errors.
2.  Having the constructor reading the parameter would make the class `WallFollowing` not be oblivious of the context in which is was constructed. Which would mean, that a (valid) `WallFollowing` object could only be created if it was startet in a ROS Launch context. It would also mean, that having more then one instance might leads to a hard to debug (and probably wrong) behavior.

Dadie:
I added the arguments to the `car-autonomous.launch` and `gazebo-car-autonomous.launch` but I'm not sure if I did it right. Can someone look over the changes in the launch file an confirm, that the changes look right?

marian42:
Looks good to me! :+1:

PR #196, opened 2019-03-13
Added <?xml ... ?> Tag in XML/Launch files where it was missing

Dadie:
* Added <?xml ... ?> Tag in XML/Launch files where it was missing
* Added UTF-8 as encoding where it was missing


Dadie:
Rebased on current master

PR #199, opened 2019-03-17
Add dms parameters to launch files

Nils-Jung:
PR #197 added some parameters for the dead man switch but forgot to include them in the main launch files. This meant that those parameters could not be set. This PR fixes this.

### DIFF
--- a/ros_ws/launch/car-autonomous.launch
+++ b/ros_ws/launch/car-autonomous.launch
@@ -3,18 +3,24 @@
 
     <arg name="marker_frame_id" default="hokuyo"/>
     <arg name="joystick_type" default="ps3"/>
+    <arg name="dms_enabled" default="true" />
+    <arg name="dms_check_rate" default="20" />
+    <arg name="dms_expiration" default="100" />
 
-    <include file="$(find car_control)/launch/car_control.launch"/>
+    <include file="$(find car_control)/launch/car_control.launch">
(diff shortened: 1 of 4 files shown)


marian42:
Why is this needed? There are already default values set in `car_control.launch`. Why do these need to be copied to the main launch file? It doesn't look like this has any effect.

Nils-Jung:
> Why is this needed? There are already default values set in `car_control.launch`. Why do these need to be copied to the main launch file? It doesn't look like this has any effect.

If a launch file includes another launch file, arguments have to be passed down for them to work. Please try the command `roslaunch launch/gazebo_car-autonomous.launch dms_enabled:=false` in both this branch and the current master branch to see the difference.
Further reference: https://wiki.ros.org/roslaunch/XML/arg#Introduction

marian42:
Ok, I understand the argument is needed to expose a command line parameter. In that case, would it make sense to leave out the `dms_check_rate` and `dms_expiration` parameters? I can't imagine a case where these would be set via comand line.

marian42:
I refactored the `dms_enable` parameter in PR #201 so this may no longer be needed.

StefanJVA:
> I refactored the `dms_enable` parameter in PR #201 so this may no longer be needed.

So can we delete this PR then?

Nils-Jung:
This branch seems to be obsolete now.

PR #200, opened 2019-03-20
Refactor emergency stop so that is independent of the lidar resolution

### LINKED ISSUE #194: Function to get LIDAR range data for a certain degree is implemented at least twice
The function 
```C++
float EmergencyStop::rangeAtDegree(const sensor_msgs::LaserScan::ConstPtr&, float)
``` 
in  `ros_ws/src/autonomous/src/emergency_stop.cpp` and the function 
```C++
float WallFollowing::getRangeAtDegree(const sensor_msgs::LaserScan::ConstPtr&, float)
```
in `ros_ws/src/autonomous/src/wall_following.cpp` implement both a function go get LIDAR range data for a certain degree. This should be refactored to make sure the range data for a certain degree is consistent between both classes.

marian42:
fixes #194

### DIFF
--- a/ros_ws/src/autonomous/include/emergency_stop.h
+++ b/ros_ws/src/autonomous/include/emergency_stop.h
@@ -12,8 +12,15 @@
 #include <ros/console.h>
 #include <ros/ros.h>
 
+constexpr float DEG_TO_RAD = M_PI / 180.0;
+
 constexpr const char* TOPIC_LASER_SCAN = "/scan";
 constexpr const char* TOPIC_EMERGENCY_STOP = "/emergency_stop";
+
+constexpr float SAMPLE_ANGLE = 6 * DEG_TO_RAD;
+constexpr float RANGE_THRESHOLD = 0.3;
+
+constexpr float MAX_RANGE = 30;
 
 class EmergencyStop
 {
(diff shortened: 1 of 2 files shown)


ros_ws/src/autonomous/include/emergency_stop.h
```diff
@@ -12,43 +12,31 @@
...
 
+constexpr float DEG_TO_RAD = M_PI / 180.0;
+
 constexpr const char* TOPIC_LASER_SCAN = "/scan";
 constexpr const char* TOPIC_EMERGENCY_STOP = "/emergency_stop";
 
+constexpr float SAMPLE_ANGLE = 6 * DEG_TO_RAD;
+constexpr float RANGE_THRESHOLD = 0.3;
+
+constexpr float MAX_RANGE = 30;
```
Dadie: Is there a reason why the max range was changed from 4 to 30?

ros_ws/src/autonomous/src/emergency_stop.cpp
```diff
@@ -2,67 +2,53 @@
...
-    // so one step equals an angle of 0,375°
-    // to access range at angle theta, you have to divide the given theta by the step size
-    float range = lidar->ranges[theta / 0.375];
-
-    // check for absurd or nan values
-    // the max range should be 4 so the car doesn't get affected by walls very far away
-    if (range > 4 || std::isinf(range))
+    int index_start = available_samples / 2 - SAMPLE_ANGLE / lidar->angle_increment / 2;
+    int index_end = available_samples / 2 + SAMPLE_ANGLE / lidar->angle_increment / 2;
+    for (int i = index_start; i < index_end; i++)
```
Dadie: The old ```rangeAtDegree``` checked for invalid indexes. I think that was quite useful, but as the lidar does not changes on runtime it should not be a normal conditional and instead it could be realized via an assertion. This way the overhead of the additional condition is only present on debug builds.

Such a assertion could look like:
```C++
ROS_ASSERT_MSG(i >= 0 & i < static_cast<int>(lidar->angle_max / lidar->angle_increment), "Lidar data are accessed with invalid index. i = %d", i);
```

ros_ws/src/autonomous/include/emergency_stop.h
marian42: This is the value we use in `wall_following.h` and is closer to the actual maximum range of the Lidar. I think it is "more correct" to detect a distant wall rather than no wall, even thouh neither would trigger an emergency brake.

Issue #190, opened 2019-03-12
Enable manual and autonomous driving with different buttons


marian42:
Currently when the Dead Man's Switch is pressed and one tries to drive manually and autonomously, the autonomous command is ignored until you stop using the keyboard/gamepad.

Instead, we want to have two separate "Dead Man's Switch" buttons.

To Do:
- Duplicate the behaviour of the Dead Man's Switch
- Send state of the second button in keyboard and gamepad controller
- Decide where the drive parameters are selected based on the current mode. This could happen in the multiplexer or in the car controller. Maybe the multiplexer is no longer needed.
- Display the state in the window of the keyboard controller
- Merge the autonomous and manual launch files for the real and virtual car

PR #201, opened 2019-03-20
Separate buttons for manual and autonomous driving

marian42:
Implements #190


marian42:
This is already implemented. If you set `mode_override` to 1 or 2, it will act like the button is always pressed.

ros_ws/src/car_control/src/dms_controller.cpp
```diff
@@ -55,6 +83,19 @@ void DMSController::configureParameters()
         expiration_ms = 100;
     }
     this->m_expiration_time = std::chrono::duration<double>(expiration_ms / 1000.0);
+
+    int mode_override_parameter;
+    private_node_handle.getParam(PARAMETER_MODE_OVERRIDE, mode_override_parameter);
+    if (mode_override_parameter < 0 || mode_override_parameter > 2)
```
Dadie: Magic numbers 0 and 2 could be replaced with ```static_cast<int>(DriveMode::LOCKED)``` and ```static_cast<int>(DriveMode::AUTONOMOUS)```

ros_ws/src/car_control/include/dms_controller.h
```diff
@@ -1,16 +1,21 @@
...
 constexpr const char* PARAMETER_DMS_EXPIRATION = "dms_expiration";
+constexpr const char* PARAMETER_MODE_OVERRIDE = "mode_override";
 
-constexpr const char* TOPIC_DMS_HEARTBEAT = "/input/dms_heartbeat";
-constexpr const char* TOPIC_UNLOCK_MOTOR = "/commands/unlock_motor";
+constexpr const char* TOPIC_HEARTBEAT_MANUAL = "/input/heartbeat_manual";
+constexpr const char* TOPIC_HEARTBEAT_AUTONOMOUS = "/input/heartbeat_autonomous";
+constexpr const char* TOPIC_DRIVE_MODE = "/commands/drive_mode";
+
+constexpr DriveMode NO_OVERRIDE = DriveMode::LOCKED;
```
Dadie: I found the usage of this `constexpr` in the source code quite confusion as I wasn't always sure whether  `NO_OVERRIDE` was a global constant, a local constant or an enumerator.

ros_ws/src/car_control/src/drive_parameters_multiplexer.cpp
```diff
@@ -41,6 +51,11 @@ void DriveParametersMultiplexer::onUpdate(DriveParametersSource* source,
     this->m_last_updated_source = source;
 }
 
+void DriveParametersMultiplexer::driveModeCallback(const std_msgs::Int32::ConstPtr& drive_mode_message)
+{
+    this->m_drive_mode = (DriveMode)drive_mode_message->data;
```
Dadie: Here you could add an assertion to make sure the given data is indeed a valid DriveMode value.

ros_ws/src/teleoperation/include/keyboard_controller.h
```diff
@@ -30,12 +32,21 @@ enum class KeyIndex : int
...
 
-constexpr int KEY_COUNT = 5;
+constexpr int KEY_COUNT = 6;
 
-constexpr std::array<Keycode, KEY_COUNT> KEY_CODES = { Keycode::W, Keycode::A, Keycode::S, Keycode::D, Keycode::SPACE };
+constexpr std::array<Keycode, KEY_COUNT> KEY_CODES = { Keycode::W, Keycode::A,     Keycode::S,
+                                                       Keycode::D, Keycode::SPACE, Keycode::B };
+
+enum class DriveMode : int
+{
```
Dadie: Would it be possible to somehow move the `enum class DriveMode` to a common library for both the *teleoperation* and *car_control* program? If this is not possible in a reasonable manner, than this suggestion could be moved in a new issue (with low priority).

ros_ws/src/teleoperation/src/keyboard_controller.cpp
```diff
@@ -181,12 +209,13 @@ void KeyboardController::publishDriveParameters()
...
-void KeyboardController::unlockMotorCallback(const std_msgs::Bool::ConstPtr& unlock_motor_message)
+void KeyboardController::driveModeCallback(const std_msgs::Int32::ConstPtr& drive_mode_message)
 {
-    if (this->m_car_unlocked != unlock_motor_message->data)
+    auto mode = (DriveMode)drive_mode_message->data;
+    if (this->m_drive_mode != mode)
     {
-        this->m_car_unlocked = unlock_motor_message->data;
-        this->updateWindowColor();
+        this->m_drive_mode = mode;
```
Dadie: Here you could also add an assertion to make sure the given data is indeed a valid DriveMode value.

ros_ws/src/car_control/src/dms_controller.cpp
```diff
@@ -2,37 +2,65 @@
...
     {
-        this->publishUnlockMotor();
+        this->publishDriveMode();
         ros::spinOnce();
         loop.sleep();
     }
 }
 
-void DMSController::publishUnlockMotor()
+DriveMode DMSController::getDriveMode()
```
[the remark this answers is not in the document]
marian42: IMO the name is ok because the function has no side effects (It doesn't change the state of anything). You can run it assuming that it does nothing other then getting a member. But if you have a good idea for an alternative name, I'll take it.

ros_ws/src/car_control/include/dms_controller.h
marian42: What do you suggest I do instead?

ros_ws/src/teleoperation/include/keyboard_controller.h
marian42: We discussed the same issue when we realized that topic names are redefined in multiple packages. We decided to duplicate the names to avoid dependencies between the packages. I think this is a similar case. cc @StefanJVA

ros_ws/src/car_control/include/drive_mode.h
```diff
@@ -0,0 +1,8 @@
+#pragma once
+
+enum class DriveMode : int
```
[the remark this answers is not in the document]
marian42: Can you try this on your end and see if you get it to work? For me, this created lots of weird compile errors.

marian42:
We tested this branch on the physical car today.

ros_ws/src/teleoperation/include/keyboard_controller.h
StefanJVA: True

ros_ws/src/car_control/src/dms_controller.cpp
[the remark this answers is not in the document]
StefanJVA: I think its okay here

ros_ws/src/car_control/include/drive_mode.h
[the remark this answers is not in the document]
StefanJVA: My compiler tells me: ISO C++ forbids forward references to 'enum' types

ros_ws/src/car_control/include/dms_controller.h
StefanJVA: We could just remove this variable and use DriveMode::LOCKED in the dms_controller.cpp directly

Issue #192, opened 2019-03-12
Automate starting the urg_node


marian42:
As explained [here](https://github.com/Autonomous-Racing-PG/ros.package/wiki/LIDAR), the `urg_node` for the LIDAR needs to be started manually when using the physical car.
This should be incorporated into the launch file.

Issue #188, opened 2019-03-12
Tweak wallfollowing parameters


marian42:
The wallfollowing works in general, but the parameters need tweaking.
- Find out if following a single wall or following both walls performs better. One of them should be removed.
- Test how fast the car can go (tweak maximum speed)
- Tweak parameters, including PID values, angles, distances
- Think about improvements to the algorithm
- Improve the LIDAR sampling strategy. Currently, for each angle only one LIDAR point is considered. Instead, we could read multiple points and use the minimum or average.

marian42:
Speed and PID values where changed in 6515e01b1faa1770321d6f59c44c7d703d1c5188 and fb3fb1d89d137c748c7bb43c832b20eab7f7258f

marian42:
Some of these were implemented in the second wallfollowing node.

PR #202, opened 2019-03-22
Car improvements

marian42:
Some minor tweaks to make using the car easier
implements #192, #188


marian42:
I created #204 for the maximum speed value.

PR #206, opened 2019-04-02
fixed implicit conversion from ‘float’ to ‘double’ warning

StefanJVA:
fixed implicit conversion from ‘float’ to ‘double’ warning.

An other option could be to change all float variables to double.

### DIFF
--- a/ros_ws/src/autonomous/src/wall.cpp
+++ b/ros_ws/src/autonomous/src/wall.cpp
@@ -10,19 +10,19 @@ Wall::Wall(float angle1, float angle2, float range1, float range2)
 
 float Wall::getAngle()
 {
-    return atan((m_range1 * (float)cos(std::abs(m_angle1 - m_angle2)) - m_range2) / m_range1 *
-                (float)sin(std::abs(m_angle1 - m_angle2)));
+    return std::atan((m_range1 * std::cos(std::abs(m_angle1 - m_angle2)) - m_range2) / m_range1 *
+                std::sin(std::abs(m_angle1 - m_angle2)));
 }
 
 float Wall::predictDistance(float distancce_to_current_position)
 {
     float angle = this->getAngle();
-    float currentWallDistance = m_range2 * (float)cos(angle);
-    return currentWallDistance + distancce_to_current_position * (float)sin(angle);
+    float currentWallDistance = m_range2 * std::cos(angle);
+    return currentWallDistance + distancce_to_current_position * std::sin(angle);
 }
 
 void Wall::draw(RvizGeometryPublisher& geometry, int id, std_msgs::ColorRGBA color)
 {
-    geometry.drawLine(id, createPoint(m_range1 * (float)cos(m_angle1), m_range1 * (float)sin(m_angle1), 0.0),
-                      createPoint(m_range2 * (float)cos(m_angle2), m_range2 * (float)sin(m_angle2), 0.0), color, 0.03);
+    geometry.drawLine(id, createPoint(m_range1 * std::cos(m_angle1), m_range1 * std::sin(m_angle1), 0.0f),
+                      createPoint(m_range2 * std::cos(m_angle2), m_range2 * std::sin(m_angle2), 0.0f), color, 0.03);
 }
--- a/ros_ws/src/autonomous/src/wall.cpp
+++ b/ros_ws/src/autonomous/src/wall.cpp
@@ -10,19 +10,19 @@ Wall::Wall(float angle1, float angle2, float range1, float range2)
 
 float Wall::getAngle()
 {
-    return atan((m_range1 * cos(std::abs(m_angle1 - m_angle2)) - m_range2) / m_range1 *
-                sin(std::abs(m_angle1 - m_angle2)));
+    return atan((m_range1 * (float)cos(std::abs(m_angle1 - m_angle2)) - m_range2) / m_range1 *
+                (float)sin(std::abs(m_angle1 - m_angle2)));
 }
 
 float Wall::predictDistance(float distancce_to_current_position)
 {
     float angle = this->getAngle();
-    float currentWallDistance = m_range2 * cos(angle);
-    return currentWallDistance + distancce_to_current_position * sin(angle);
+    float currentWallDistance = m_range2 * (float)cos(angle);
+    return currentWallDistance + distancce_to_current_position * (float)sin(angle);
 }
 
 void Wall::draw(RvizGeometryPublisher& geometry, int id, std_msgs::ColorRGBA color)
 {
-    geometry.drawLine(id, createPoint(m_range1 * cos(m_angle1), m_range1 * sin(m_angle1), 0.0),
-                      createPoint(m_range2 * cos(m_angle2), m_range2 * sin(m_angle2), 0.0), color, 0.03);
+    geometry.drawLine(id, createPoint(m_range1 * (float)cos(m_angle1), m_range1 * (float)sin(m_angle1), 0.0),
+                      createPoint(m_range2 * (float)cos(m_angle2), m_range2 * (float)sin(m_angle2), 0.0), color, 0.03);
 }
--- a/ros_ws/src/autonomous/src/wall_following.cpp
+++ b/ros_ws/src/autonomous/src/wall_following.cpp
@@ -76,8 +76,8 @@ void WallFollowing::followSingleWall(const sensor_msgs::LaserScan::ConstPtr& lid
                                     createColor(1, 0, 0, 1), 0.03);
     float wallAngle = wall.getAngle();
     this->m_debug_geometry.drawLine(2, createPoint(PREDICTION_DISTANCE, -error * leftRightSign, 0),
-                                    createPoint(PREDICTION_DISTANCE + (float)cos(wallAngle) * 2,
-                                                (-error + (float)sin(wallAngle) * 2) * leftRightSign, 0),
+                                    createPoint(PREDICTION_DISTANCE + std::cos(wallAngle) * 2,
+                                                (-error + std::sin(wallAngle) * 2) * leftRightSign, 0),
                                     createColor(0, 1, 1, 1), 0.03);
 
     this->publishDriveParameters(velocity, steeringAngle);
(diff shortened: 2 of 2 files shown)


ros_ws/src/autonomous/src/wall_following.cpp
```diff
@@ -76,8 +76,8 @@ void WallFollowing::followSingleWall(const sensor_msgs::LaserScan::ConstPtr& lid
                                     createColor(1, 0, 0, 1), 0.03);
     float wallAngle = wall.getAngle();
     this->m_debug_geometry.drawLine(2, createPoint(PREDICTION_DISTANCE, -error * leftRightSign, 0),
-                                    createPoint(PREDICTION_DISTANCE + cos(wallAngle) * 2,
-                                                (-error + sin(wallAngle) * 2) * leftRightSign, 0),
+                                    createPoint(PREDICTION_DISTANCE + (float)cos(wallAngle) * 2,
+                                                (-error + (float)sin(wallAngle) * 2) * leftRightSign, 0),
```
marian42: I don't understand why this is needed. Since `wallAngle` is already a float, shouldn't `sin(wallAngle)` also be a float? Is this related to the literal `2`? Does using `2f` behave differently`?

ros_ws/src/autonomous/src/wall_following.cpp
```diff
@@ -92,7 +92,7 @@ void WallFollowing::followWalls(const sensor_msgs::LaserScan::ConstPtr& lidar)
     float correction = this->m_pid_controller.updateAndGetCorrection(error, TIME_BETWEEN_SCANS);
 
     float steeringAngle = atan(correction) * 2 / M_PI;
-    float velocity = WALL_FOLLOWING_MAX_SPEED * (1 - std::max(0.0, std::abs(steeringAngle) - 0.15));
+    float velocity = WALL_FOLLOWING_MAX_SPEED * (1 - (float)std::max(0.0, (double)std::abs(steeringAngle) - 0.15));
```
marian42: Again, I don't understand this. Is there a reasoning why a conversion to double is needed? It looks to me like all numbers in this line are floats. Can this be done without any conversions?

ros_ws/src/autonomous/src/wall_following.cpp
StefanJVA: Well when i look at this: http://www.cplusplus.com/reference/cmath/sin/
it seems like you are right for c++11. However i always get a warning that says "implicit conversion from ‘float’ to ‘double’". Maybe this is an compiler issue?

StefanJVA: @Dadie we need your c++ wizard skillz :D

Dadie: In this case the calls to `sin` and `cos` are correctly typed `double`, as those functions are defined only for `double` in the C-Header `math.h` which is included by default by the C++ Wrapper `cmath`.

To call the C++ versions either add the namespace `std::` or ~~enrich~~ pollute the current namespace with `using std;` (**please** don't).

e.g.
```C++
constexpr float CONST_SOMETHING = 1337;
auto should_be_float = std::sin(2.0f) * 2 + CONST_SOMETHING;
```

ros_ws/src/autonomous/src/wall.cpp
```diff
@@ -10,19 +10,19 @@ Wall::Wall(float angle1, float angle2, float range1, float range2)
 
 float Wall::getAngle()
 {
-    return atan((m_range1 * cos(std::abs(m_angle1 - m_angle2)) - m_range2) / m_range1 *
-                sin(std::abs(m_angle1 - m_angle2)));
+    return atan((m_range1 * (float)cos(std::abs(m_angle1 - m_angle2)) - m_range2) / m_range1 *
```
Dadie:
```suggestion
    return std::atan((m_range1 * std::cos(std::abs(m_angle1 - m_angle2)) - m_range2) / m_range1 *
```

ros_ws/src/autonomous/src/wall.cpp
```diff
@@ -10,19 +10,19 @@ Wall::Wall(float angle1, float angle2, float range1, float range2)
...
+                (float)sin(std::abs(m_angle1 - m_angle2)));
```
Dadie:
```suggestion
                std::sin(std::abs(m_angle1 - m_angle2)));
```

ros_ws/src/autonomous/src/wall.cpp
Dadie: Because `atan(double)` from `math.h` is used, by replacing it with the C++ one from `cmath` (which is defined for `float`) no conversion from `double` to `float` should happen.

ros_ws/src/autonomous/src/wall.cpp
Dadie: Because `sin(double)` from `math.h` is used, by replacing it with the C++ one from `cmath` (which is defined for `float`) no conversion from `double` to `float` should happen.

ros_ws/src/autonomous/src/wall.cpp
```diff
@@ -10,19 +10,19 @@ Wall::Wall(float angle1, float angle2, float range1, float range2)
...
 }
 
 float Wall::predictDistance(float distancce_to_current_position)
 {
     float angle = this->getAngle();
-    float currentWallDistance = m_range2 * cos(angle);
-    return currentWallDistance + distancce_to_current_position * sin(angle);
+    float currentWallDistance = m_range2 * (float)cos(angle);
```
Dadie:
```suggestion
    float currentWallDistance = m_range2 * std::cos(angle);
```
Because `cos(double)` from `math.h` is used, by replacing it with the C++ one from `cmath` (which is defined for `float`) no conversion from `double` to `float` should happen.

ros_ws/src/autonomous/src/wall.cpp
```diff
@@ -10,19 +10,19 @@ Wall::Wall(float angle1, float angle2, float range1, float range2)
...
+    return currentWallDistance + distancce_to_current_position * (float)sin(angle);
```
Dadie:
```suggestion
    return currentWallDistance + distancce_to_current_position * std::sin(angle);
```
Because `sin(double)` from `math.h` is used, by replacing it with the C++ one from `cmath` (which is defined for `float`) no conversion from `double` to `float` should happen.

ros_ws/src/autonomous/src/wall.cpp
```diff
@@ -10,19 +10,19 @@ Wall::Wall(float angle1, float angle2, float range1, float range2)
...
 }
 
 void Wall::draw(RvizGeometryPublisher& geometry, int id, std_msgs::ColorRGBA color)
 {
-    geometry.drawLine(id, createPoint(m_range1 * cos(m_angle1), m_range1 * sin(m_angle1), 0.0),
-                      createPoint(m_range2 * cos(m_angle2), m_range2 * sin(m_angle2), 0.0), color, 0.03);
+    geometry.drawLine(id, createPoint(m_range1 * (float)cos(m_angle1), m_range1 * (float)sin(m_angle1), 0.0),
```
Dadie:
```suggestion
    geometry.drawLine(id, createPoint(m_range1 * std::cos(m_angle1), m_range1 * std::sin(m_angle1), 0.0f),
```
Because `cos(double)` from `math.h` is used, by replacing it with the C++ one from `cmath` (which is defined for `float`) no conversion from `double` to `float` should happen. Also added the `float` literal to the floating number.

ros_ws/src/autonomous/src/wall.cpp
```diff
@@ -10,19 +10,19 @@ Wall::Wall(float angle1, float angle2, float range1, float range2)
...
+                      createPoint(m_range2 * (float)cos(m_angle2), m_range2 * (float)sin(m_angle2), 0.0), color, 0.03);
```
Dadie:
```suggestion
                      createPoint(m_range2 * std::cos(m_angle2), m_range2 * std::sin(m_angle2), 0.0f), color, 0.03);
```
Because `cos(double)` and `sin(double)` from `math.h` is used, by replacing it with the C++ one from `cmath` (which is defined for `float`) no conversion from `double` to `float` should happen. Also added the `float` literal to the floating number.

ros_ws/src/autonomous/src/wall_following.cpp
```diff
@@ -76,8 +76,8 @@ void WallFollowing::followSingleWall(const sensor_msgs::LaserScan::ConstPtr& lid
                                     createColor(1, 0, 0, 1), 0.03);
     float wallAngle = wall.getAngle();
     this->m_debug_geometry.drawLine(2, createPoint(PREDICTION_DISTANCE, -error * leftRightSign, 0),
-                                    createPoint(PREDICTION_DISTANCE + cos(wallAngle) * 2,
-                                                (-error + sin(wallAngle) * 2) * leftRightSign, 0),
+                                    createPoint(PREDICTION_DISTANCE + (float)cos(wallAngle) * 2,
```
Dadie:
```suggestion
                                    createPoint(PREDICTION_DISTANCE + std::cos(wallAngle) * 2,
```
Because `cos(double)` from `math.h` is used, by replacing it with the C++ one from `cmath` (which is defined for `float`) no conversion from `double` to `float` should happen.

ros_ws/src/autonomous/src/wall_following.cpp
Dadie:
```suggestion
    float velocity = WALL_FOLLOWING_MAX_SPEED * (1 - std::max(0.0f, std::abs(steeringAngle) - 0.15f));
```
Added the `float` literal to the floating number, which should fix the warning.

ros_ws/src/autonomous/src/wall_following.cpp
Dadie:
```suggestion
                                                (-error + std::sin(wallAngle) * 2) * leftRightSign, 0),
```
Because `sin(double)` from `math.h` is used, by replacing it with the C++ one from `cmath` (which is defined for `float`) no conversion from `double` to `float` should happen.

PR #207, opened 2019-04-02
Removed unneeded dependency that caused a warning

StefanJVA:
Fixes the warning 

"WARNING: package "racer_description" should not depend on metapackage "ros_controllers" but on its packages instead"

that appeared each time the workspace was build from scratch.

### DIFF
--- a/ros_ws/src/simulation/racer_description/package.xml
+++ b/ros_ws/src/simulation/racer_description/package.xml
@@ -13,6 +13,5 @@
 
   <!-- Dependencies -->
   <depend>catkin</depend>
-  <depend>ros_controllers</depend>
 
 </package>


StefanJVA:
> I know for a fact that the racer_description package needs _some_ kind of ROS controller package, I just don't know which. The dependency on the metapackage is mainly just for rosdep since the required packages are not installed by default. If you check out this branch, everything seems to be fine since the required packages are already installed on your system; the only way to be sure whether any dependency is actually required or not is to reinstall ROS from scratch. Even the Travis build may succeed but actually running Gazebo will fail.
> I've already (unsuccessfully) tried before to fix this problem (#148) but I figured it is too much work for too little payoff.

Oki Doki, then let us close this PR and forget that it ever happened :D

PR #208, opened 2019-04-03
Fit the new emergency stop to Marians Lidar calculations. 

malikatamne:
The new emergency stop range is dependent on the cars width.

### DIFF
--- a/ros_ws/src/autonomous/src/emergency_stop.cpp
+++ b/ros_ws/src/autonomous/src/emergency_stop.cpp
@@ -13,7 +13,7 @@ bool EmergencyStop::emergencyStop(const sensor_msgs::LaserScan::ConstPtr& lidar)
 
     // calculate the sample_angle of the lidar to be used to check if there is
     // an obstacle in front of the car, with respect to the RANGE_THRESHOLD
-    float sample_angle = (360 * CAR_BUMPER_LENGTH) / (M_PI * RANGE_THRESHOLD * 2) * DEG_TO_RAD;
+    float sample_angle = atan(CAR_BUMPER_LENGTH / 2 / RANGE_THRESHOLD);
 
     int index_start = available_samples / 2 - sample_angle / lidar->angle_increment / 2;
     int index_end = available_samples / 2 + sample_angle / lidar->angle_increment / 2;
@@ -24,16 +24,7 @@ bool EmergencyStop::emergencyStop(const sensor_msgs::LaserScan::ConstPtr& lidar)
     // Instead of calculating the range average, we are more cautious because we would 
     // rather have false positives instead of false negatives.
     // i.e. we would rather stop too much than crash into an obstacle.
-    float min_dist = MAX_RANGE;
-    for (int i = index_start; i < index_end; i++)
-    {
-        float range = lidar->ranges[i];
-
-        if(min_dist > range)
-        {
-            min_dist = range;
-        }
-    }
+    auto min_dist = std::min(MAX_RANGE, *std::min_element(lidar->ranges.begin() + index_start, lidar->ranges.begin() + index_end));
 
     return min_dist < RANGE_THRESHOLD;
 }
--- a/ros_ws/src/autonomous/src/emergency_stop.cpp
+++ b/ros_ws/src/autonomous/src/emergency_stop.cpp
@@ -9,11 +9,16 @@ EmergencyStop::EmergencyStop()
 
 bool EmergencyStop::emergencyStop(const sensor_msgs::LaserScan::ConstPtr& lidar)
 {
+    ROS_ASSERT_MSG(RANGE_THRESHOLD <= MAX_RANGE, "Threshold is bigger then max range. Function will always return false.");
+    ROS_ASSERT_MSG(MAX_RANGE > 0, "Max range is zero or below. Function will mostly return true.");
+    ROS_ASSERT_MSG(RANGE_THRESHOLD > 0, "Threshold is zero or below. Function will mostly return false.");
+
     int available_samples = (lidar->angle_max - lidar->angle_min) / lidar->angle_increment;
 
     // calculate the sample_angle of the lidar to be used to check if there is
     // an obstacle in front of the car, with respect to the RANGE_THRESHOLD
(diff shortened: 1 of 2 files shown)


ros_ws/src/autonomous/src/emergency_stop.cpp
```diff
@@ -10,32 +10,32 @@ EmergencyStop::EmergencyStop()
...
 {
     int available_samples = (lidar->angle_max - lidar->angle_min) / lidar->angle_increment;
-    int samples_used = 0;
-    float range_sum = 0;
 
-    int index_start = available_samples / 2 - SAMPLE_ANGLE / lidar->angle_increment / 2;
-    int index_end = available_samples / 2 + SAMPLE_ANGLE / lidar->angle_increment / 2;
+    // calculate the sample_angle of the lidar to be used to check if there is
+    // an obstacle in front of the car, with respect to the RANGE_THRESHOLD
+    float sample_angle = (360 * CAR_BUMPER_LENGTH) / (M_PI * RANGE_THRESHOLD * 2) * DEG_TO_RAD;
```
marian42: What is the reasoning behind calculating the `sample_angle` like this? I expected something like `atan(CAR_BUMPER_LENGTH / 2 / RANGE_THRESHOLD)`.
Also, I think that the comment above the line isn't needed, since it explains what is done, but not why.

ros_ws/src/autonomous/src/emergency_stop.cpp
```diff
@@ -10,32 +10,32 @@ EmergencyStop::EmergencyStop()
...
+    // Instead of calculating the range average, we are more cautious because we would 
+    // rather have false positives instead of false negatives.
+    // i.e. we would rather stop too much than crash into an obstacle.
+    float min_dist = MAX_RANGE;
     for (int i = index_start; i < index_end; i++)
     {
         float range = lidar->ranges[i];
 
-        if (range < MAX_RANGE && !std::isinf(range))
+        if(min_dist > range)
```
marian42: Please run the autoformatting so that the space after `if` gets back in. Consider renaming `min_dist` to `min_range` or `min_distance`.

ros_ws/src/autonomous/src/emergency_stop.cpp
malikatamne: I wanted the arc length of the lidar to be the same as the width of the car. Since the arc length is dependent on the arc angle (sample_angle) and the radius (RANGE_THRESHOLD), I used this [formula](https://theengineeringmindset.com/arc-length-how-to-calculate/) to determine the exact angle, at which the arc length will be the same as the width of the car, with a radius of RANGE_THRESHOLD.

ros_ws/src/autonomous/src/emergency_stop.cpp
```diff
@@ -10,32 +10,32 @@ EmergencyStop::EmergencyStop()
...
     ROS_ASSERT_MSG(index_start >= 0 && index_end < available_samples,
-                   "SAMPLE_ANGLE is too big. Trying to access lidar samples out of bounds.");
+                   "sample_angle is too big. Trying to access lidar samples out of bounds.");  
 
+    // determine the minimum distance between the car and a potetial obstacle
+    // Instead of calculating the range average, we are more cautious because we would 
+    // rather have false positives instead of false negatives.
+    // i.e. we would rather stop too much than crash into an obstacle.
+    float min_dist = MAX_RANGE;
     for (int i = index_start; i < index_end; i++)
```
Dadie: If all you're doing is finding the smalles number in the array `lidar->ranges` you could use 
[`std::min_element`](https://en.cppreference.com/w/cpp/algorithm/min_element) from `<algorithm>` to reduce the LoC.

C++11 Example
```C++
auto min_dist = std::min(MAX_RANGE, std::min_element(lidar->ranges + index_start, lidar->ranges + index_end));
```
C++17 Example (using the Parallelism TS)
```C++
auto  min_dist = std::min(std::execution::par, MAX_RANGE, std::min_element(lidar->ranges + index_start, lidar->ranges + index_end));
```

ros_ws/src/autonomous/src/emergency_stop.cpp
```diff
@@ -10,32 +10,23 @@ EmergencyStop::EmergencyStop()
...
-    }
+    // determine the minimum distance between the car and a potetial obstacle
+    // Instead of calculating the range average, we are more cautious because we would 
+    // rather have false positives instead of false negatives.
+    // i.e. we would rather stop too much than crash into an obstacle.
+    auto min_dist = std::min(MAX_RANGE, *std::min_element(lidar->ranges.begin() + index_start, lidar->ranges.begin() + index_end));
 
-    float range_average = range_sum / samples_used;
-    return range_average < RANGE_THRESHOLD;
+    return min_dist < RANGE_THRESHOLD;
```
Dadie: Here additional checks could be added to make sure the calculations always make sense
```C++
    ROS_ASSERT_MSG(RANGE_THRESHOLD <= MAX_RANGE, "Threshold is bigger then max range. Function will always return false.");
    ROS_ASSERT_MSG(MAX_RANGE > 0, "Max range is zero or below. Function will mostly return true.");
    ROS_ASSERT_MSG(RANGE_THRESHOLD > 0, "Threshold is zero or below. Function will mostly return false.");
    ROS_ASSERT_MSG(min_dist >= 0, "The minimal distance between the car and a potential obstacle is below zero.");
```

ros_ws/src/autonomous/src/emergency_stop.cpp
```diff
@@ -10,32 +10,23 @@ EmergencyStop::EmergencyStop()
 bool EmergencyStop::emergencyStop(const sensor_msgs::LaserScan::ConstPtr& lidar)
 {
     int available_samples = (lidar->angle_max - lidar->angle_min) / lidar->angle_increment;
```
Dadie: Because `lidar->ranges` is a `std::vector` this calculation could be replaced by simply calling the `.size()` function of `lidar->ranges`
```suggestion
    size_t available_samples = lidar->ranges.size();
```

StefanJVA:
@Dadie @marian42 @malikatamne whats up with this PR? :D
All requested changes are outdated, however no one has merged it.

PR #212, opened 2019-04-14
Add tf messages in vesc_sim

Nils-Jung:


ros_ws/src/simulation/vesc_sim/src/vesc_sim.cpp
```diff
@@ -21,6 +22,7 @@ VESCSimulator::VESCSimulator()
     m_base_frame = "base_link";
 
     this->m_odometry_publisher = this->m_node_handle.advertise<nav_msgs::Odometry>("odom", 10);
+    m_tf_publisher.reset(new tf::TransformBroadcaster);
```
marian42: Does this create a memory leak?

ros_ws/src/simulation/vesc_sim/include/vesc_sim.h
```diff
@@ -21,6 +23,7 @@ class VESCSimulator
     private:
     ros::NodeHandle m_node_handle;
     ros::Publisher m_odometry_publisher;
+    boost::shared_ptr<tf::TransformBroadcaster> m_tf_publisher;
```
Dadie: Is there a reason why you are using a `shared_ptr`? As far as I can tell the `shared_ptr` does not seem necessary (as the object is not shared between other objects or functions).

Dadie:
As the function `timerCallback` got longer I'm wondering if it is possible that more then one instance of the function can run at the same time (as they are called external via a timer event). Does anyone knows if that could happen?

ros_ws/src/simulation/vesc_sim/include/vesc_sim.h
Nils-Jung: This code was... uh... _inspired_ by the vesc_to_odom node in the vesc_ackermann package. There is no reason why a `shared_ptr` is used here, I just know that it works. If you want to make some improvements, feel free to push anything onto the branch

Nils-Jung:
Since we are using `ros::spin()` the whole node is single-threaded, so there shouldn't be any callbacks running at the same time.

ros_ws/src/simulation/vesc_sim/include/vesc_sim.h
Dadie: I removed the `shared_ptr` and it seems to work, but please tests it also on your computer as I'd like to prevent "but it worked on my computer!!11"-Problems ^^,

Nils-Jung: It seems to work just fine 👍

PR #218, opened 2019-04-22
Circle-based wallfollowing

marian42:
A second approach for wallfollowing that fits circles into the laser scan and steers the car based on these circles


ros_ws/src/autonomous/wallfollowing2/script/wallfollowing.py
```diff
@@ -0,0 +1,153 @@
...
+
+    show_circle_in_rviz(left_circle, left_wall, 0)
+    show_circle_in_rviz(right_circle, right_wall, 1)
+
+
+laser_scan = None
+
+rospy.Subscriber("/scan", LaserScan, laser_callback)
+drive_parameters_publisher = rospy.Publisher(
+    "/input/drive_param/autonomous", drive_param, queue_size=1)
```
StefanJVA: You could move this string to the to of the file and call the variable TOPIC_DRIVE_PARAMETERS (just like in wall_following.h in wallfollowing1). However that's not really a must.

PR #226, opened 2019-05-05
Update wallfollowing parameters

marian42:


ros_ws/src/autonomous/wallfollowing2/script/wallfollowing.py
```diff
@@ -17,17 +17,21 @@
...
-ERROR_DEAD_ZONE = 0.2
+ERROR_SPEED_DECREASE = 4
+ERROR_DEAD_ZONE = 0.15
 
 UPDATE_FREQUENCY = 60
 
+MAX_ACCELERATION = 0.5
+
+last_speed = 0
+
```
Dadie: If those values are not constants (as they are changed) could they be move to the `.launch` file as args?

ros_ws/src/autonomous/wallfollowing2/script/wallfollowing.py
```diff
@@ -90,7 +100,9 @@ def find_left_right_border(points, margin_relative=0.1):
 
 
 def follow_walls(left_circle, right_circle):
-    predicted_car_position = Point(0, 1.5)
+    global last_speed
+
+    predicted_car_position = Point(0, 1.4 + last_speed)
```
Dadie: May give the magic number `0.14` a name or add a comment describing what the magic number does.

ros_ws/src/autonomous/wallfollowing2/script/wallfollowing.py
```diff
@@ -149,7 +166,7 @@ def handle_scan():
 
 timer = rospy.Rate(UPDATE_FREQUENCY)
 
-pid = PIDController(1.5, 0.2, 0.01)
+pid = PIDController(1.5, 0.2, 0.02)
```
Dadie: May add the P,I,D factors to the `.launch` args.

PR #234, opened 2019-05-08
Add crash detector node

marian42:
- The `lap_timer` package was renamed to `simulation_tools` to better reflect what it does (Since implementing the lap timer, other tools were added to the package)
- Added the crash detector node from the `marian-q` branch because it is needed for other projects


marian42:
The loading time issue has been fixed!

PR #238, opened 2019-05-13
Changed Some Parameters

NklasF:


ros_ws/src/navigation_stack/navigation_stack_config/config/amcl_params.yaml
```diff
@@ -4,19 +4,19 @@ odom_frame_id: "odom"
 #base_frame_id: "base_footprint"
 global_frame_id: "map"
 
-odom_model_type: "diff"
-odom_alpha5: 0.1
+odom_model_type: "diff-corrected"
```
Nils-Jung: These changes aren't really necessary because we aren't using AMCL anymore but I guess it's still good to include them.

ros_ws/src/navigation_stack/navigation_stack_config/config/costmap_common_params.yaml
```diff
@@ -8,9 +8,9 @@ obstacle_layer:
  enabled: true
  obstacle_range: 3.0
  raytrace_range: 3.5
- inflation_radius: 0.2
```
Nils-Jung: Because the inflation radius is missing the global planner creates a path much too close to the wall. This might be problematic with an actual car; driving without a crash should be the highest priority.

PR #239, opened 2019-05-17
Sami fann

Sami19944:
New:
- ai_driver node that uses a neural network to drive the car
- ai_trainer node that is used to train neural networks using an evolutionary genetic algorithm (still needs to be improved as it is self implemented)

How to start:
roslaunch launch/training_environment.launch
roslaunch launch/neuralnetwork.launch

- change training parameters in "ros.package/ros_ws/src/autonomous/neuralnetwork/launch/neuralnetwork.launch"
- change the nn_config_folder parameter to your path to "ros.package/ros_ws/src/autonomous/neuralnetwork/config"
- check reading/writing permissions for the nn_config_folder

Also:
the includes are still getting switched, thats why it cant compile :(


marian42:
Can you disable clang-format where it should not change the include order? ([see here for how to do it](https://github.com/Autonomous-Racing-PG/ros.package/blob/master/CPP_STYLE_GUIDE.md#formatting))

Sami19944:
i think travis cant compile because the fann lib is missing on the server

marian42:
If the library can be installed via apt, please add it at line 21 in `.travis.yml`.

ros_ws/src/autonomous/neuralnetwork/include/ai_vector_math.h
```diff
@@ -0,0 +1,84 @@
+#pragma once
+
+#include <algorithm>
+#include <cstdlib>
+#include <vector>
+
+inline void add(std::vector<fann_type>& a, std::vector<fann_type>& b)
+{
```
Dadie: You could add a assertion here to make sure that `b` has at least the same size as `a` e.g.
```C++
ROS_ASSERT_MSG(b.size() >= a.size(), "...");
```

ros_ws/src/autonomous/neuralnetwork/include/ai_vector_math.h
```diff
@@ -0,0 +1,84 @@
...
+    for (uint i = 0; i < a.size(); i++)
+    {
+        a[i] = a[i] + b;
+    }
+}
+
+inline void mult(std::vector<fann_type>& a, std::vector<fann_type>& b)
+{
+    for (uint i = 0; i < a.size(); i++)
```
Dadie: You could add a assertion here to make sure that `b` has at least the same size as `a` e.g.
```C++
ROS_ASSERT_MSG(b.size() >= a.size(), "...");
```

ros_ws/src/autonomous/neuralnetwork/src/ai_trainer.cpp
```diff
@@ -0,0 +1,246 @@
...
+    FANN::connection connections[size];
+    net->get_connection_array(connections);
+    std::vector<unsigned int> from_neurons;
+    std::vector<unsigned int> to_neurons;
+    std::vector<float> weights;
+    for (int i = 0; i < size; i++)
+    {
+        weights.push_back(connections[i].weight);
+    }
+    message.weights = weights;
```
Dadie: You could fill the `message.weights` vector directly which would prevent unnecessary copies e.g.
```C++
    message.weights.clear();
    for (size_t i = 0; i < size; i++)
    {
        message.weights.push_back(connections[i].weight);
    }
```

PR #240, opened 2019-05-17
Add bigger track

Nils-Jung:
This PR adds a scaled version of our simulation track with a scale factor of 2.5. To use this bigger track, add `world:=racetrack_decorated_2_big` to any (main) launch file invocation.
The reason why this is interesting is because it shows how some algorithms deal with bigger tracks than our current, comparably thin track. For example, the current wallfollowing algorithm crashes halfway through the bigger track.
This new track is missing support from most of the helper scripts from the simulation_tools package. Feel free to add to this PR if anyone wants to work on that.


marian42:
Are there any changes in the model files other than the scale? If they are just scaled, you can reuse the old files like this:
```
<geometry>
    <mesh><uri>model://walls.dae</uri>
    <scale>2.5 2.5 2.5</scale></mesh>
</geometry>```

Nils-Jung:
I had searched for an option to just apply a scale but couldn't find any. This makes it much easier 😅

PR #242, opened 2019-05-19
Added dynamic configuration to teleoperation, emergency stop and wallfollowing1

Dadie:
As we discussed, I tired to add dynamic configuration to the teleoperation package. I move some constants like ```STEERING_SPEED``` to the resp. class and declared them as normal members which allows us to change them via dynamic configuration.

### DIFF
--- a/ros_ws/src/autonomous/emergency_stop/CMakeLists.txt
+++ b/ros_ws/src/autonomous/emergency_stop/CMakeLists.txt
@@ -9,11 +9,20 @@ find_package(catkin REQUIRED COMPONENTS
   drive_msgs
   roscpp
   std_msgs
+  dynamic_reconfigure
 )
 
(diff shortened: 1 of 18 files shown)


marian42:
I like the idea of dynamic reconfiguration at runtime! But I think it should be limited to parameters where a realistic usecase exists in which it could be helpful to change parameters at runtime. In my opinion, it would be useful to change all kinds of parameters related to the vehicle's performance that need to be tweaked continuously (maximum speed, wallfollwoing parameters, emergency braking thresholds etc.). However, I think it's not useful to change key or joystick axis assginments at runtime (vs restarting ROS for that). At the same time, changing these makes the code more cluttered.

Dadie:
added dynamic config for "emergency stop" and "wallfollowing1"

PR #243, opened 2019-05-20
Run autopep8 over the codebase

Dadie:


Dadie:
Used wrong version of autopep8, can be closed.

PR #244, opened 2019-05-20
Add lean topic

Nils-Jung:
The topic "/input/lean" goes from `0.0` (ram into left wall) to `1.0` (ram into right wall). For safety reasons, the input for this topic should be reasonably close to 0.5.

### DIFF
--- a/ros_ws/src/autonomous/wallfollowing2/script/wallfollowing.py
+++ b/ros_ws/src/autonomous/wallfollowing2/script/wallfollowing.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 
 import rospy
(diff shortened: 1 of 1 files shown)


marian42:
I think it would be better to calculate meaninful values for the lean parameter in the wallfollowing script instead of receiving them via a ros topic. For example, you could determine if you're in a left or right curve from the x value of the circle center and lean accordingly.

Nils-Jung:
> I think it would be better to calculate meaningful values for the lean parameter in the wallfollowing script instead of receiving them via a ros topic. For example, you could determine if you're in a left or right curve from the x value of the circle center and lean accordingly.

Makes sense, I close this PR then.

Issue #221, opened 2019-04-28
Use ros time instead of system time


marian42:
Since the simulation can be sped up or slowed down in gazebo, all timing in the project should be done using ros time (wich is the simulated time in gazebo) instead of the system time.

PR #245, opened 2019-05-20
Increases the dms_expiration time

Luc4sW:
- Fixes keypress stutter for slower processors


marian42:
The time should be independent of the processor speed. I think this parameter should not be changed before #221 is fixed, since that ticket probably solves the problem.

StefanJVA:
This should be fixed with PR #252

PR #248, opened 2019-05-22
Using dms_controller for emergency stop

StefanJVA:
The dms_controller now listens for messages in topic "/input/emergencystop".
If a message with an timestamp arrives on this topic, the dms controller locks itself for (default) 3000ms.

To test this you have to switch to wallfollowing1 inside the gazebo.launch , since wallfollowing2 has no emergencystop right now.

### DIFF
--- a/ros_ws/src/car_control/include/dms_controller.h
+++ b/ros_ws/src/car_control/include/dms_controller.h
@@ -41,7 +41,7 @@ class DMSController
     /**
      * @brief How long the dead man's switch blocks all inputs, after receving an emergency stop message
      */
-    std::chrono::duration<double> m_emergencystop_exploration_time;
+    std::chrono::duration<double> m_emergencystop_expiration_time;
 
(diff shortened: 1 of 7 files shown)


ros_ws/src/car_control/include/dms_controller.h
```diff
@@ -36,19 +38,26 @@ class DMSController
      * @brief How old the last dead man's switch heartbeat can be, in ms
      */
     std::chrono::duration<double> m_expiration_time;
+    /**
+     * @brief How long the dead man's switch blocks all inputs, after receving an emergency stop message
+     */
+    std::chrono::duration<double> m_emergencystop_exploration_time;
```
marian42: Should this be expiration instead of exploration?

marian42:
I think the emergency stop node could be started in `gazebo.launch` instead of `autonomous_driving.launch` so that it is run regardless of the autonomous controller.

Issue #232, opened 2019-05-06
Add emergency stop to wall_following2


Nils-Jung:

marian42:
I think it would make sense to handle emergency stop in a similar way as the Dead Man's Switch currently works. We could have multiple publishers that send an emergency stop message and then stop regardless of the current autonomous controller. This could be done in the multiplexer or in the dms_controller.

marian42:
For example, the Q value in Q learning is a predictor for how long it will take until the car crashes. The q_learning node could publish an emergency stop message once the Q value drops below a threshold.

PR #249, opened 2019-05-22
Stefan emergencystop separate

StefanJVA:
Put emergency_stop into separate package and start it in gazebo.launch .
This means that the emergency_stop is active by default when gazebo.launch is started.
This fixes ticket #232 .
Note that is might interfere with qlearning and other approaches where an emergency stop is not desired.


ros_ws/src/autonomous/emergency_stop/launch/emergency_stop.launch
```diff
@@ -0,0 +1,14 @@
...
+    <!-- emergency_stop node -->
+    <node
+      respawn="true"
+      pkg="emergency_stop"
+      type="emergency_stop"
+      name="emergency_stop"
+      output="screen" >
+    </node>
+
+</launch>
```
marian42: This might be done in `gazebo.launch` instead in its own file. The screen output might not be necesary.

ros_ws/src/autonomous/emergency_stop/CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
...
+
+## Compile as C++11, supported in ROS Kinetic and newer
+add_compile_options(-std=c++11)
+
+## Find catkin macros and libraries
+find_package(catkin REQUIRED COMPONENTS
+  drive_msgs
+  roscpp
+  rospy
+  std_msgs
```
marian42: I think rospy is not needed in this package.

ros_ws/src/autonomous/emergency_stop/package.xml
```diff
@@ -0,0 +1,25 @@
...
+  <!-- License -->
+  <license>MIT</license>
+  <license>GPLv3</license>
+
+  <!-- Dependencies -->
+  <buildtool_depend>catkin</buildtool_depend>
+
+  <depend>drive_msgs</depend>
+  <depend>roscpp</depend>
+  <depend>rospy</depend>
```
marian42: Not needed.

ros_ws/src/autonomous/emergency_stop/launch/emergency_stop.launch
StefanJVA: Well it is like that in every other autonomous package as well, so why is this different to wallfollowing or q_learning?

PR #251, opened 2019-05-22
Add particle filter

Nils-Jung:
Replaces the rather slow AMCL localization with rather fast particle filter localization.


ros_ws/launch/navigation_stack.launch
```diff
@@ -10,6 +10,9 @@
     <arg name="use_gpu" default="true"/>
     <arg name="joystick_type" default="xbox360"/>
     <arg name="visualize_lidar" default="false"/>
+    <arg name="pf_angle_step" default="18"/>
+	<arg name="pf_max_particles" default="4000"/>
```
StefanJVA: Maybe that a bit high? I get similar good results with much lower values.

Nils-Jung:
>     * Is there a way to include this as a git-submodule instead of coping the code?

Not really. I had to make some changes to the code to make it work for our car and I don't think a submodule supports that.

>     * How do we make sure that it is easy to install the needed python dependencies? I remember that it was not really strait forward.

I guess it would be possible to write a script for that but that would downgrade pip without people realizing beforehand. I'd rather like to err on the safe side and just point them to the existing tutorial, I'll copy it into our readme (with the spelling fix).

ros_ws/launch/navigation_stack.launch
Nils-Jung: It's just the default value from the `particle_filter` package. Any suggestions what our default should be?

PR #254, opened 2019-05-26
Car odometry

Nils-Jung:
Adds/Readds odometry for the actual car.

Notable changes:
- Removed the (outdated) vesc submodule and added the updated vesc package from https://github.uconn.edu/ret15108/vesc
- Added a params_to_yaml rosnode to convert the car_config header into a .yaml file because ROS can only read .yaml files (maybe the parameters should be saved as a yaml and converted into a header instead?)
- Added travis check for consistency between yaml and header file

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,3 @@
 [submodule "ros_ws/src/external_packages/vesc"]
 	path = ros_ws/src/external_packages/vesc
-	url = https://github.com/mit-racecar/vesc.git
+	url = https://github.uconn.edu/ret15108/vesc
--- a/scripts/travis/check-consistency.sh
+++ b/scripts/travis/check-consistency.sh
@@ -5,5 +5,5 @@ source $pathofscript/../../ros_ws/devel/setup.bash
 
 mv $pathofvescsim/config/car_config.yaml $pathofvescsim/config/car_config_copy.yaml || { echo 'Could not find "car_config.yaml" in package "vesc_sim".' ; exit 1; }
 rosrun vesc_sim params_to_yaml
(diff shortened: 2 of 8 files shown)


.gitmodules
```diff
@@ -1,3 +0,0 @@
-[submodule "ros_ws/src/external_packages/vesc"]
-	path = ros_ws/src/external_packages/vesc
-	url = https://github.com/mit-racecar/vesc.git
```
marian42: Can you just change this URL to https://github.uconn.edu/ret15108/vesc instead of copying the code?

scripts/travis/check-consistency.sh
```diff
@@ -0,0 +1,9 @@
+pathofscript=$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )
+pathofvescsim=$pathofscript/../../ros_ws/src/simulation/vesc_sim
+
+source $pathofscript/../../ros_ws/devel/setup.bash
+
+mv $pathofvescsim/config/car_config.yaml $pathofvescsim/config/car_config_copy.yaml || { echo 'Could not find "car_config.yaml" in package "vesc_sim".' ; exit 1; }
+rosrun vesc_sim params_to_yaml
+cmp -s $pathofvescsim/config/car_config.yaml $pathofvescsim/config/car_config_copy.yaml || { echo 'Inconsisteny between "vesc_sim/include/car_config.h" and "vesc_sim/config/car_config.yaml" detected. Please run "rosrun vesc_sim params_to_yaml" to fix this.' ; exit 0; }
```
marian42: Shouldn't this exit with 1?

.gitmodules
Nils-Jung: ...Pretty much. I forgot that the new one should be a submodule as well :/

scripts/travis/check-consistency.sh
Nils-Jung: Yup. Nice catch.

PR #255, opened 2019-05-28
Various fixes

Nils-Jung:
Added robot_description & robot_state_publisher to car.launch
Added cartographer parameter for easier access
Fixed cartographer scripts

### DIFF
--- a/ros_ws/launch/car.launch
+++ b/ros_ws/launch/car.launch
@@ -2,7 +2,7 @@
 <launch>
     <arg name="joystick_type" default="ps3"/>
 
-    <param name="robot_description" command="$(find xacro)/xacro --inorder '$(find racer_description)/urdf/racer.xacro' use_gpu:=true visualize_lidar:=true laser_sample_count:=1080"/>
+    <param name="robot_description" command="$(find xacro)/xacro --inorder '$(find racer_description)/urdf/racer.xacro'"/>
 
     <include file="$(find car_control)/launch/car_control.launch"/>
 
--- a/ros_ws/launch/car.launch
+++ b/ros_ws/launch/car.launch
@@ -2,6 +2,8 @@
 <launch>
     <arg name="joystick_type" default="ps3"/>
 
+    <param name="robot_description" command="$(find xacro)/xacro --inorder '$(find racer_description)/urdf/racer.xacro' use_gpu:=true visualize_lidar:=true laser_sample_count:=1080"/>
+
     <include file="$(find car_control)/launch/car_control.launch"/>
 
     <include file="$(find teleoperation)/launch/remote_control.launch">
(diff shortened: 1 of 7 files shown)


ros_ws/launch/car.launch
```diff
@@ -2,6 +2,8 @@
 <launch>
     <arg name="joystick_type" default="ps3"/>
 
+    <param name="robot_description" command="$(find xacro)/xacro --inorder '$(find racer_description)/urdf/racer.xacro' use_gpu:=true visualize_lidar:=true laser_sample_count:=1080"/>
```
marian42: `use_gpu:=true visualize_lidar:=true laser_sample_count:=1080"`
Why are these needed? Shouldn't they be irrelevant outside of Gazebo?

Nils-Jung: They are but xacro needs _some_ values for those to compile. Maybe it's possible to write some default values into the xacro file, idk.

Nils-Jung: Fixed it. :P

ros_ws/src/simulation/racer_description/urdf/racer.xacro
```diff
@@ -3,6 +3,10 @@
 <robot name="myrobot" 
   xmlns:xacro="http://www.ros.org/wiki/xacro">
 
+  <xacro:arg name="use_gpu" default="true"/>
+  <xacro:arg name="visualize_lidar" default="false"/>
+  <xacro:arg name="laser_sample_count" default="1080"/>
```
marian42: Since these are only used in `racer_plugins.gazebo`, the default values should be set there.

ros_ws/src/simulation/racer_description/urdf/racer.xacro
```diff
@@ -3,6 +3,10 @@
...
+
   <xacro:include filename="$(find racer_description)/urdf/racer_plugins.gazebo" />
```
marian42: Now that this xacro file is used outside of gazebo, I think it makes sense to make this include optional.
```xml
<xacro:arg name="use_gazebo" default="false"/>
<xacro:if value="$(arg use_gazebo)">
    <xacro:include filename="$(find racer_description)/urdf/racer_plugins.gazebo" />
</xacro:if>
```

Nils-Jung: Does it matter though? The content of racer_plugins is only used by gazebo anyway, making an argument to include that file is pretty much useless.

ros_ws/src/simulation/racer_description/urdf/racer.xacro
Nils-Jung: Done.

PR #257, opened 2019-06-01
Added ros message publisher to lap_timer

Nils-Jung:


ros_ws/src/simulation/simulation_tools/src/simulation_tools/lap_timer.py
```diff
@@ -34,6 +35,8 @@ def __init__(self, name, checkpoints):
         self.next_checkpoint = 0
         self.history = []
         self.start = None
+        self.msg_seq = 0
+        self.pub = rospy.Publisher("/lap_time", DurationStamped, queue_size=1)
```
marian42: Why do you use a DurationStamped instead of `std_msgs/Duration Message`?
Does `marti_common_messages` need to be installed manually?
The variable names should not be abbreviations, please use something like `message_index` and `lap_time_publisher`.

ros_ws/src/simulation/simulation_tools/src/simulation_tools/lap_timer.py
```diff
@@ -61,6 +64,12 @@ def complete_lap(self):
             rospy.loginfo("Lap " + str(len(self.history)) + " (" + self.name + "): " +  # nopep8
                           format_duration(duration) + ", average: " + format_duration(average))  # nopep8
         self.start = time
+        self.msg_seq += 1
+        pub_msg = DurationStamped()
```
marian42: Consider renaming this to `message` or `duration_message`.

ros_ws/src/simulation/simulation_tools/src/simulation_tools/lap_timer.py
```diff
@@ -61,6 +64,12 @@ def complete_lap(self):
...
+        pub_msg.header.stamp = time
+        pub_msg.header.seq = self.msg_seq
```
marian42: This value is set automatically by ros, you don't need to set it.

ros_ws/src/simulation/simulation_tools/src/simulation_tools/lap_timer.py
Nils-Jung: Done.

ros_ws/src/simulation/simulation_tools/src/simulation_tools/lap_timer.py
Nils-Jung: Thanks, good to know.

ros_ws/src/simulation/simulation_tools/src/simulation_tools/lap_timer.py
Nils-Jung: I figured the increasing sequence numbers might be useful for some applications down the line and the overhead doesn't seem that big to me. Unfortunately, `marti_common_messages` would need to be installed manually. @Sami19944 , would such a message header be useful for you?

Sami19944: i cant think of a scenario where i can use them.  important are the lap times at the end of a finished lap

Sami19944:
approved, but cant find the button

Nils-Jung:
> approved, but cant find the button

Files changed > Review changes ;)

PR #259, opened 2019-06-02
multiple cars in gazebo

Sami19944:
spawn and controll multiple cars in gazebo


StefanJVA:
@Sami19944 how do i test this?

Sami19944:
@StefanJVA this is still a work in progress.
If you start the training_enviroment.launch every gazebo part of a car will have its own prefix for its topic.
TODO:
Each car must get its own car_controller and vesc node

PR #267, opened 2019-06-04
Navigation stack

Nils-Jung:
The navigation stack uses a particle filter (= SLAM) and the TEB local planner to localize the robot on a predetermined map and calculate a path to a given goal.

The navigation stack is pretty much feature complete now, so I think it's time to merge it into the master.

To use it, first launch `launch/navigation_stack.launch` and set the initial postition of the car in rviz using `2D Pose Estimate` in the top bar. Then set a goal for the car to drive towards using `2D Nav Goal`. Note that for both the pose estimate and the nav goal pressing the mouse button sets the position and releasing the mouse button sets the direction.

I had to merge the master branch into the navigation stack branch to get it up to speed but that introduced a few bugs. I tried to fix them but some might have slipped under my radar. Please help me test this branch to see if everything (including non-navigation-stack-related tasks) still works.


marian42:
Is it possible to add the particle filter as a git submodule?

Nils-Jung:
> Is it possible to add the particle filter as a git submodule?

Not really. I had to make some changes to the particle filter code and afaik submodules don't support that.

marian42:
It would be possible by creating a fork and adding that fork as a submodule.

Nils-Jung:
> It would be possible by creating a fork and adding that fork as a submodule.

Good idea. I see what I can do.

Nils-Jung:
The particle_filter package now resides as a fork in https://github.com/Autonomous-Racing-PG/particle_filter. You probably have to use the command `git submodule update --init --recursive` to get the package onto your machine.

marian42:
I like the solution. It might be worth rewriting the commit message of 7736279459f660e731a4274911b735b947b594cd so that it't easier to see why we did the fork and what is has changed.

PR #271, opened 2019-06-09
Refactor Q-Learning code

marian42:
Some changes to the Q-Learning code that make it more readable and more maintainable
- Update every time a lidar scan is received, instead of using a timer
- Reuse more code between the training and the driving node
- Only use the taken actions for the neural net loss function, instead of all of them
- Improve time measurement and debug output

This was developed together with improvements to make the car drive faster with Q-learning. Since that doesn't work yet, I'm submitting only the refactoring part as a pull request for now.


marian42:
The travis error seems to be unrelated to this pullrequest...

marian42:
Weird. I just tried again and it can drive most of the track after about 180 episodes. It reached 500 steps for the first time at episode 146.

Issue #47, opened 2018-11-02
Setup IMU


FlyingdutchmanC:
- [x] ros interaction, receiving data

marian42:
The IMU is set up and can publish data to a ROS topic. I created #279 for the calibration.

PR #275, opened 2019-06-10
Add a node for the Razor IMU to the car launch file

marian42:
refs #47


Nils-Jung:
- Has this been tested on the hardware yet? I didn't follow the progress for the IMU very closely but as far as I could tell it didn't work on any of our weekly meetings.
- Does this introduce any dependencies? If so, they should be included somewhere. Maybe there should be a metapackage in our repo which doesn't contain any code but lists all dependencies not included in any other packages (https://wiki.ros.org/Metapackages).

marian42:
- It has been tested with the real IMU on my Ubuntu machine. There could still be differences to the car setup.
- I added a package with a dependency to the razor_imu_9dof package. I moved the configuration file into that package but that means it can't be a metapackage now that it contains a file.

marian42:
The current version was successfully tested on the car.

ros_ws/src/hardware/CMakeLists.txt
```diff
@@ -0,0 +1,4 @@
+cmake_minimum_required(VERSION 2.8.3)
+project(hardware)
+find_package(catkin REQUIRED)
+catkin_metapackage()
```
Nils-Jung: You said that this package cannot be a metapackage because it contains a parameter file. Should this line then be here?

PR #277, opened 2019-06-14
Improve Q-Learning for slow speeds

marian42:
Still no velocity control unfortunately, but faster training and an action to drive straight.


marian42:
It is intentional that the epsilon greedy strategy does not go below 30% random, however this PR introduces a behaviour where only every second episode uses the epsilon greedy strategy, while the others are pure exploitation (0% random).

I think it would be good if others test the Q-Learning code to see which of us is the outlier. Even better if we can get it to work on your machine.

PR #280, opened 2019-06-16
fixed #266 by adding dynamic reconfiguration to WallFollowing2

### LINKED ISSUE #266: Add dynamic reconfiguration for wallfollowing2

Dadie:


ros_ws/src/autonomous/wallfollowing2/script/wallfollowing.py
```diff
@@ -207,6 +211,53 @@ def laser_callback(scan_message):
...
+        'straight_smoothing': config.straight_smoothing,
+        'barrier_size_realtive': config.barrier_size_realtive,
+        'barrier_lower_limit': config.barrier_lower_limit,
+        'barrier_upper_limit': config.barrier_upper_limit,
+        'barrier_exponent': config.barrier_exponent,
+        'controller_p': config.controller_p,
+        'controller_i': config.controller_i,
+        'controller_d': config.controller_d}
+    parameters = Parameters(new_parameters)
+    parameters.load()
```
marian42: This loads the parameters from ROS. It should work as expected if you remove this line.

ros_ws/src/autonomous/wallfollowing2/script/wallfollowing.py
```diff
@@ -207,6 +211,53 @@ def laser_callback(scan_message):
...
+        'max_acceleration': config.max_acceleration,
+        'corner_cutting': config.corner_cutting,
+        'straight_smoothing': config.straight_smoothing,
+        'barrier_size_realtive': config.barrier_size_realtive,
+        'barrier_lower_limit': config.barrier_lower_limit,
+        'barrier_upper_limit': config.barrier_upper_limit,
+        'barrier_exponent': config.barrier_exponent,
+        'controller_p': config.controller_p,
+        'controller_i': config.controller_i,
+        'controller_d': config.controller_d}
```
marian42: The creation of `new_parameters` could be rewritten like this:
`new_parameters = {key: getattr(config, key) for key in DEFAULT_PARAMETERS}`

ros_ws/src/autonomous/wallfollowing2/script/wallfollowing.py
```diff
@@ -207,6 +211,53 @@ def laser_callback(scan_message):
...
+          max_acceleration: {max_acceleration},
+          corner_cutting: {corner_cutting},
+          straight_smoothing: {straight_smoothing},
+          barrier_size_realtive: {barrier_size_realtive},
+          barrier_size_realtive: {barrier_lower_limit},
+          barrier_upper_limit: {barrier_upper_limit},
+          barrier_exponent: {barrier_exponent},
+          controller_p: {controller_p},
+          controller_i: {controller_i},
+          controller_d: {controller_d}
```
marian42: Consider implementing a `__str__` method for the parameters class. It could look like this:
```python
def __str__(self):
    return '\n'.join(name + ": " + str(getattr(self, name)) for name in self.names)
```

After creating the new parameters object, you can create the log message like this:
```python
rospy.loginfo('Reconfigure Request:\n' + str(parameters))
```

ros_ws/src/autonomous/wallfollowing2/script/wallfollowing.py
```diff
@@ -219,5 +270,7 @@ def laser_callback(scan_message):
 drive_parameters_publisher = rospy.Publisher(
     TOPIC_DRIVE_PARAMETERS, drive_param, queue_size=1)
 
+dyn_cfg_srv = Server(wallfollowing2Config, dyn_cfg_callback)
```
marian42: You don't need to store `dyn_cfg_srv` since you're not using it.

ros_ws/src/autonomous/wallfollowing2/script/wallfollowing.py
```diff
@@ -207,6 +211,53 @@ def laser_callback(scan_message):
     last_scan = scan_time
 
 
+def dyn_cfg_callback(config, level):
```
marian42: This function should be renamed to avoid abbreviations.

ros_ws/src/autonomous/wallfollowing2/script/wallfollowing.py
```diff
@@ -207,6 +211,53 @@ def laser_callback(scan_message):
...
+        'barrier_exponent': config.barrier_exponent,
+        'controller_p': config.controller_p,
+        'controller_i': config.controller_i,
+        'controller_d': config.controller_d}
+    parameters = Parameters(new_parameters)
+    parameters.load()
+    pid.p = parameters.controller_p
+    pid.i = parameters.controller_i
+    pid.d = parameters.controller_d
+    return config
```
marian42: Is the return value used anywhere?

marian42:
You could move the `Parameters` class and the `DEFAULT_PARAMETERS` into a separate file and reference it from wallfollowing2.cfg to avoid redundancy. But that might be overkill.

ros_ws/src/autonomous/wallfollowing2/script/wallfollowing.py
Dadie: The example from the [dynamic_reconfigure tutorial](https://wiki.ros.org/dynamic_reconfigure/Tutorials/SettingUpDynamicReconfigureForANode%28python%29) returned the config. As I've no idea how and in what kind of context the callback is called, I've no idea what removing that return implies ^^,

Dadie: According to the error message I get after removing the return I assume, that the config has to be returned.

PR #282, opened 2019-06-17
Lucas emergency stop

Luc4sW:
-The warning concerning emergency status will be issued only once
-Added visualization for min and actual range
-Added ros parameter for min and max range

### DIFF
--- a/ros_ws/src/autonomous/emergency_stop/src/emergency_stop.cpp
+++ b/ros_ws/src/autonomous/emergency_stop/src/emergency_stop.cpp
@@ -1,7 +1,9 @@
 #include "emergency_stop.h"
 #include <std_msgs/Time.h>
 
-EmergencyStop::EmergencyStop():m_emergency_status(EmergencyStatus::UNUSED), m_debug_geometry(this->m_node_handle, TOPIC_VISUALIZATION, LIDAR_FRAME)
+EmergencyStop::EmergencyStop()
+    : m_emergency_status(EmergencyStatus::UNUSED)
+    , m_debug_geometry(this->m_node_handle, TOPIC_VISUALIZATION, LIDAR_FRAME)
 {
     this->m_lidar_subscriber =
         this->m_node_handle.subscribe<sensor_msgs::LaserScan>(TOPIC_LASER_SCAN, 1, &EmergencyStop::lidarCallback, this);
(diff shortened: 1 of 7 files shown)


ros_ws/src/autonomous/emergency_stop/src/emergency_stop.cpp
```diff
@@ -2,54 +2,114 @@
...
+    if ((this->m_range_threshold > 2.f) || (this->m_range_threshold < 0.3f))
+    {
+        ROS_WARN_STREAM("m_range_threshold is set to "
+                        << this->m_range_threshold << ". This value is currently limited to (0.3, 2). Using default: "
+                        << RANGE_THRESHOLD_DEFAULT << ".");
+        this->m_range_threshold = RANGE_THRESHOLD_DEFAULT;
+    }
+
+    private_node_handle.getParam(MAX_RANGE, this->m_max_range);
+    if ((this->m_max_range > 30.f) || (this->m_max_range <= 2.f))
```
Nils-Jung: If 2 should be included and 30 excluded, then the `>` should be `>=` and vice versa.
```suggestion
    if ((this->m_max_range >= 30.f) || (this->m_max_range < 2.f))
```

ros_ws/src/autonomous/emergency_stop/src/emergency_stop.cpp
```diff
@@ -2,54 +2,114 @@
...
+    this->m_emergency_stop_publisher = this->m_node_handle.advertise<std_msgs::Time>(TOPIC_EMERGENCY_STOP, 1);
+    this->configureParameters();
+}
+
+void EmergencyStop::configureParameters()
+{
+    ros::NodeHandle private_node_handle("~");
+
+    private_node_handle.getParam(RANGE_THRESHOLD, this->m_range_threshold);
+    if ((this->m_range_threshold > 2.f) || (this->m_range_threshold < 0.3f))
```
Nils-Jung: If 2 and 0.3 should be excluded, then the `>` should be `>=` and vice versa. But wouldn't it be fine if they were both included?
```suggestion
    if ((this->m_range_threshold >= 2.f) || (this->m_range_threshold <= 0.3f))
```

ros_ws/src/autonomous/emergency_stop/src/emergency_stop.cpp
Luc4sW: Yes. After looking over it again, i think it should also be fine to include 30 since this value was our default max value before.

ros_ws/src/autonomous/emergency_stop/src/emergency_stop.cpp
Luc4sW: Yes. Regarding your question, i think it would be fine to include 2, but 0.3 is so close to the car, it wouldn't really make sense as a min value for the emergency stop.

PR #285, opened 2019-06-19
Various q_learning fixes

Nils-Jung:
- Added model save files to gitignore
- Moved launch of q_learning nodes into separate launch file
- Added pytorch dependency to readme 
- Added "use_gpu" argument support to launch file


ros_ws/src/autonomous/q_learning/launch/autonomous_driving.launch
```diff
@@ -1,11 +0,0 @@
...
-<launch>
-<!-- Launches all necessary nodes that let the car drive autonomously. -->
-    <node
-      respawn="true"
-      pkg="q_learning"
-      type="drive.py"
-      name="q_learning_driver"
-      output="screen" >
-    </node>
-</launch>
```
marian42: Why was this deleted?

ros_ws/launch/qlearning.launch
```diff
@@ -21,19 +22,5 @@
...
-    output="screen" >
-  </node>
-
-  <node
-      pkg="simulation_tools"
-      type="crash_detector"
-      name="crash_detector"
-      output="screen" >
-  </node>
+  <include file="$(find q_learning)/launch/q_learning.launch"/>
```
marian42: Why did you move these?

ros_ws/src/autonomous/q_learning/launch/autonomous_driving.launch
Nils-Jung: I thought this was a file used for debugging purposes during development that erroneously made it's way onto the master. Sorry, I'll restore it.

ros_ws/launch/qlearning.launch
Nils-Jung: I personally think it is better to have more modular launch files, I think it helps code readability. If you don't agree though or if it messes with your workflow I can revert it of course.

PR #287, opened 2019-06-20
Added basic reward plotter for qlearning

Nils-Jung:
Right now the plotter only plots the last 50 reward values. If anything else should be plotted, let me know.

### DIFF
--- a/ros_ws/src/autonomous/q_learning/scripts/plotter.py
+++ b/ros_ws/src/autonomous/q_learning/scripts/plotter.py
@@ -1,0 +1,37 @@
+#!/usr/bin/env python
+
+from PyQt4 import QtCore, QtGui
+from parameters import *
+import pyqtgraph as pg
+import rospy
+import sys
+import numpy
+from collections import deque
+from std_msgs.msg import Float32
+
+
+def update_reward_list(msg):
+    reward_list.append(msg.data)
+
+
+def update_plot():
+    reward_plot.plot(reward_list, clear=True)
+
+
+reward_list = deque(maxlen=50)
+
+app = QtGui.QApplication(sys.argv)
+
+reward_plot = pg.plot()
+
+update_plot_timer = QtCore.QTimer()
+update_plot_timer.timeout.connect(update_plot)
+update_plot_timer.start(500)
+
+rospy.init_node('q_learning_plotter', anonymous=True)
+rospy.Subscriber(TOPIC_REWARDS, Float32, update_reward_list)
+
+if (sys.flags.interactive != 1) or not hasattr(QtCore, 'PYQT_VERSION'):
+    app.exec_()
+
+rospy.spin()
--- a/ros_ws/src/autonomous/q_learning/scripts/plotter.py
+++ b/ros_ws/src/autonomous/q_learning/scripts/plotter.py
@@ -5,31 +5,39 @@ from parameters import *
 import pyqtgraph as pg
 import rospy
 import sys
-import numpy
-from collections import deque
-from std_msgs.msg import Float32
+from q_learning.msg import EpisodeResult
 
 
-def update_reward_list(msg):
-    reward_list.append(msg.data)
+def update_result_lists(msg):
+    reward_list.append(msg.reward)
+    length_list.append(msg.length)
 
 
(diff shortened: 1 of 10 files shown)


marian42:
It would also be useful to plot the episode length.

Nils-Jung:
I've added a plot for the episode length but I'm not sure how useful it is. The length is pretty much the same as the reward, so you basically got the same plot twice :P

marian42:
It is useful to have a comparable value if you experiment with different reward functions.

Nils-Jung:
Anything else that should be plotted? Otherwise the branch is ready to be reviewed/merged.

ros_ws/src/autonomous/q_learning/scripts/plotter.py
```diff
@@ -0,0 +1,55 @@
...
+    length_list.append(msg.length)
+
+
+def update_plot():
+    reward_plot.setData(range(len(reward_list))[-50:], reward_list[-50:])
+    length_plot.setData(range(len(length_list))[-50:], length_list[-50:])
+
+
+reward_list = list()
+length_list = list()
```
marian42: You could use a `deque` instead of a list here to limit memory usage.

ros_ws/src/autonomous/q_learning/scripts/plotter.py
```diff
@@ -0,0 +1,55 @@
...
+
+
+def update_result_lists(msg):
+    reward_list.append(msg.reward)
+    length_list.append(msg.length)
+
+
+def update_plot():
+    reward_plot.setData(range(len(reward_list))[-50:], reward_list[-50:])
+    length_plot.setData(range(len(length_list))[-50:], length_list[-50:])
```
marian42: Have you considered removing the timer and updating the plot in `EpisodeResult` callback?

marian42: I would create a constant for the number of episodes and also increase it to 200 or 400.

ros_ws/src/autonomous/q_learning/scripts/plotter.py
```diff
@@ -0,0 +1,55 @@
...
+length_plot = main_plot.plot(
+    pen='g',
+    symbol='o',
+    symbolPen='w',
+    symbolBrush='g',
+    symbolSize=8,
+    name='length')
+reward_plot = main_plot.plot(
+    pen='r',
+    symbol='o',
```
marian42: The symbols are not needed IMO.

ros_ws/src/autonomous/q_learning/scripts/plotter.py
```diff
@@ -0,0 +1,55 @@
+#!/usr/bin/env python
+
+from PyQt4 import QtCore, QtGui
+from parameters import *
+import pyqtgraph as pg
```
marian42: Please add a note to the readme on how to install pyqt4 and pyqtgraph.
I just installed them like this:
```
sudo apt install python-qt4
sudo python -m pip install pyqtgraph
```

Nils-Jung: I've already added `python-pyqtgraph` as an apt package which also includes qt4.

ros_ws/src/autonomous/q_learning/scripts/plotter.py
Nils-Jung: > Have you considered removing the timer and updating the plot in the `EpisodeResult` callback?

I swear I've tried something like this before and it didn't work because the graph cannot be updated by a different thread. Works now, though.

ros_ws/src/autonomous/q_learning/scripts/plotter.py
```diff
@@ -29,25 +36,20 @@ def update_plot():
...
 reward_plot = main_plot.plot(
     pen='r',
-    symbol='o',
-    symbolPen='w',
-    symbolBrush='r',
-    symbolSize=8,
+    # symbol='o',
+    # symbolPen='w',
+    # symbolBrush='r',
+    # symbolSize=8,
```
marian42: These should be removed.

Issue #265, opened 2019-06-03
Make simulation_tools work with the big racetrack/multiple tracks


Nils-Jung:

Nils-Jung:
Implemented in #288 .

PR #288, opened 2019-06-22
Added simulation_tools support for racetrack_decorated_2_big

Nils-Jung:
implements #265


ros_ws/src/simulation/simulation_tools/src/simulation_tools/track.py
```diff
@@ -134,4 +142,7 @@ def get_position(self, total_distance, distance_to_center=0):
     (-6.00, -0.17),
     (2.64, -0.36)))
 
+if world_name == "racetrack_decorated_2_big":
+    POINTS = POINTS * 2.5
```
marian42:
```suggestion
    POINTS *= 2.5
```
This changes the data in place, instead of creating a copy.

PR #289, opened 2019-06-22
Added various emergency stop functionality

Luc4sW:
-The warning concerning emergency status will be issued only once
-Added visualization for min and actual range
-Added  "TODO Tooltip" messages in the configuration of emergency stop 
-Added min, max values in the configuration of emergency stop

### DIFF
--- a/ros_ws/src/autonomous/emergency_stop/CMakeLists.txt
+++ b/ros_ws/src/autonomous/emergency_stop/CMakeLists.txt
@@ -45,7 +45,8 @@ include_directories(
 )
 
 # emergency stop node
-add_executable(emergency_stop src/emergency_stop.cpp)
+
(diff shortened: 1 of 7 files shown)


Luc4sW:
What should have changed in the visualization?  That's exactly the same thing.  The big yellow line is from the visualization of the wallfollowing, something seems to be going awry.  The visualization of the Emergency Stop is not activated, you have set the checkmark in Rviz.

StefanJVA:
> What should have changed in the visualization? That's exactly the same thing. The big yellow line is from the visualization of the wallfollowing, something seems to be going awry. The visualization of the Emergency Stop is not activated, you have set the checkmark in Rviz.

Jep you are totally right. My fold. :)

PR #295, opened 2019-07-05
Readme overhaul

Nils-Jung:
This PR is a major overhaul of our readme aimed to increase readability and usability for external readers. As most changes in this PR are of subjective nature it would be nice to leave some comments/suggestions.


marian42:
IMO the headline and introduction should make it easier to understand what this project offers to people who are not familiar with it. They might be more interested in our capabilities than the fact that it belongs to TU Dortmund. An example for a title could be "Autonomous Racing Software Stack and Simulation Tools".
I think we should also mention that we focus on LIDAR-based driving.

It would be useful to have a list of features and a screenshot at the top of the readme file, but this could be done in another PR.

Consider naming the top level headlines "Installation" and "Usage".

Consider adding a `#` to the explanatory texts inside bash code to make it clear that they should not be executed.

PR #302, opened 2019-07-21
Malik policy gradient2

malikatamne:
A new method for autonomous driving using the policy gradient approach.


ros_ws/launch/qlearning.launch
```diff
@@ -15,7 +15,8 @@
       <arg name="verbose" value="false"/>
       <arg name="use_gpu" value="$(arg use_gpu)"/>
       <arg name="extra_gazebo_args" value="--profile qlearning" unless="$(arg realtime_simulation)"/>
-      <arg name="laser_sample_count" value="32"/>
+      <arg name="laser_sample_count" value="8"/>
+      <arg name="car_highlighting" value="$(arg car_highlighting)"/>
```
StefanJVA: This throws an error since the argument is no longer present

PR #303, opened 2019-07-24
Policy Gradient method for autonomous driving

malikatamne:
Created a new launch file for the policy gradient approach. Formatted the code.


ros_ws/src/autonomous/q_learning/scripts/parameters.py
```diff
@@ -16,17 +17,54 @@
 
 # General parameters
 
-ACTIONS = [(-0.6, 0.2), (0.6, 0.2), (0, 0.2)]
+ACTIONS = [(-0.8, 0.1), (0.8, 0.1), (0.5, 0.2),
+           (-0.5, 0.2), (0, 0.2), (0, 0.4)]
 ACTION_COUNT = len(ACTIONS)
 
 # Only use some of the LIDAR measurements
 # When changing this value, also update laser_sample_count in qlearning.launch
```
StefanJVA: As this comment states, the laser_sample_count should be changed inside qlearning.launch
The value in qlearning.launch is still 32, while the value in policy_gradient.launch is 8.
However i tested the qlearning.launch and it does still work. So i am not sure if changing the variable is necessary. Maybe @marian42 knows more?

marian42:
The way I understand it, Policy Gradient is a reinforcment learning approach, but not Q-Learning, since it doesn't estimate a Q value. In this PR, the policy gradient code is inside the q_learning package. Consider making a copy of the `q_learning` directory that is called `policy_gradient`. That way we can keep them separate and they'll work independently. We could then, in a separate PR, create another package for shared code.

StefanJVA:
> The way I understand it, Policy Gradient is a reinforcment learning approach, but not Q-Learning, since it doesn't estimate a Q value. In this PR, the policy gradient code is inside the q_learning package. Consider making a copy of the `q_learning` directory that is called `policy_gradient`. That way we can keep them separate and they'll work independently. We could then, in a separate PR, create another package for shared code.
> Alternatively, we could keep them in the same directory and rename `q_learning` to `reinforcement_learning`.

Since separating the 2 packages would require a fair amount of work i would suggest that we rename q_learning into reinforcement_learning as @marian42 suggests. What do you think @malikatamne ?

ros_ws/src/autonomous/q_learning/scripts/parameters.py
malikatamne: I actually don't know why this argument is needed. It seems to work fine even when the argument is deleted from the launch file. The reason why I did it that way was because I used Marians Q-Learning code as a template to write my Policy-Gradient code.

malikatamne:
> > The way I understand it, Policy Gradient is a reinforcment learning approach, but not Q-Learning, since it doesn't estimate a Q value. In this PR, the policy gradient code is inside the q_learning package. Consider making a copy of the `q_learning` directory that is called `policy_gradient`. That way we can keep them separate and they'll work independently. We could then, in a separate PR, create another package for shared code.
> > Alternatively, we could keep them in the same directory and rename `q_learning` to `reinforcement_learning`.
> 
> Since separating the 2 packages would require a fair amount of work i would suggest that we rename q_learning into reinforcement_learning as @marian42 suggests. What do you think @malikatamne ?

Yes you are right @marian42, I will rename the directory `q_learning` to `reinforcement_learning`.

ros_ws/src/autonomous/q_learning/scripts/parameters.py
marian42: The launch file argument specifies the resolution of the simulated Lidar. The default value, without an argument is 1080. `LASER_SAMPLE_COUNT` specifies how many laser samples are used for the autonomous driving model (currently 8). The code will downsample the laser scan message if it contains more points than needed. The reason to change the launch file argument is to save computation time by not simulating the scan points that will be discarded.

malikatamne: This makes sense, thank you for the explanation.

PR #305, opened 2019-07-29
Lucas urg node delay

Luc4sW:
The delay is not necessary. The parameter name for the ip address was wrong.


marian42:
This could be squashed since the delay doesn't need to be in the repository.

PR #307, opened 2019-07-29
Remove car_tf

marian42:
Removes the car_tf package since it is not used anywhere (and wasn't used in a long time).


StefanJVA:
I am not really sure if this is necessary. The code is at least functional and could theoretically be used in the future.
If we decide to rearrange the repo we could put this code into a separate repo, while not including it into the .rosinstall

PR #310, opened 2019-08-04
Chase Cam

marian42:
A program that updates the Gazebo client camera so that it follows the car.

Known problem: Sometimes Gazebo will stop updating the car position if the chase cam is running.


marian42:
This seems to be a bug in Gazebo, I happens for me as well, as mentioned above. As a work around, I restart Gazebo again and again until it works. :frowning_face:

StefanJVA:
> This seems to be a bug in Gazebo, it happens for me as well, as mentioned above. As a work around, I restart Gazebo again and again until it works. ☹️

Mhh okay thats not a really nice solution. However since this is more of a "fun feature" i guess i will merge it.

Issue #318, opened 2019-08-12
Make Q-Learning and Policy Gradient use different parameters


marian42:
Right now, they use the same `ACTIONS` parameter, but they should have separate ones

PR #325, opened 2019-08-16
Refactor Reinforcement Learning

marian42:
- Add two separate parameter files for Q-Learning and Policy Gradient (#318)
- Create a common superclass for Q-Learning and Policy Gradient to reduce duplicated code
- Make Policy Gradient simpler and more efficient
- Make variable naming consistent across Q-Learning and Policy Gradient, don't use "q_learning" where it refers to both approaches
- Add Policy Gradient Driving Node
- Enable learning progress plot for Policy Gradient


ros_ws/src/autonomous/reinforcement_learning/scripts/parameters_policy_gradient.py
```diff
@@ -0,0 +1,55 @@
...
+            self.load_state_dict(torch.load(MODEL_FILENAME))
+            rospy.loginfo("Model parameters loaded.")
+
+    def save(self):
+        torch.save(self.state_dict(), MODEL_FILENAME)
+
+
+# Start by loading previously trained parameters.
+# If this is False, training will start from scratch
+CONTINUE = False
```
StefanJVA: Why are some parameters at the start of the file and some at the end? Maybe this should consistent. (Unless there is some reason for this of course :D )

ros_ws/src/autonomous/reinforcement_learning/scripts/parameters_q_learning.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def save(self):
+        torch.save(self.state_dict(), MODEL_FILENAME)
+
+
+# Training parameters
+
+# Start by loading previously trained parameters.
+# If this is False, training will start from scratch
+CONTINUE = False
```
StefanJVA: Same here. Some Parameters at the top and some at the bottom.

ros_ws/src/autonomous/reinforcement_learning/scripts/training_node.py
```diff
@@ -0,0 +1,160 @@
...
+                self.on_complete_episode()
+        else:
+            self.state = new_state
+            self.action = self.select_action(new_state)
+            self.perform_action(self.action)
+            self.episode_length += 1
+            self.total_step_count += 1
+
+    def on_complete_step(self, state, action, reward, next_state):
+        pass
```
StefanJVA: Thats a pretty nice method :D

ros_ws/src/autonomous/reinforcement_learning/scripts/drive_policy_gradient.py
```diff
@@ -0,0 +1,45 @@
...
+
+from reinforcement_learning_node import ReinforcementLearningNode
+import os
+import rospy
+from parameters_policy_gradient import Policy, ACTIONS, LASER_SAMPLE_COUNT
+import torch
+from torch.distributions import Categorical
+
+
+class QLearningDrivingNode(ReinforcementLearningNode):
```
StefanJVA: Why is this called "QLearningDrivingNode" when its purpose is to drive with policygradiant?

ros_ws/src/autonomous/reinforcement_learning/scripts/training_node.py
marian42: It's like an abstract method, not sure if this is the correct way to do it. It would also be possible to raise an exception ("not implemented").

PR #326, opened 2019-08-17
Added launch file for real car navstack

### LINKED ISSUE #321: Create a launch file for the navigation stack on the physical car

Nils-Jung:
fixes #321


marian42:
The list of launch files (line 80) should also be updated.

PR #331, opened 2019-08-25
Updated emergency_stop

### LINKED ISSUE #328: Enable steering during emergency stop
It should be possible to steer thte car while emergency stop is active.

Nils-Jung:
fixes #328

### DIFF
--- a/ros_ws/src/car_control/src/dms_controller.cpp
+++ b/ros_ws/src/car_control/src/dms_controller.cpp
@@ -64,7 +64,8 @@ bool DMSController::getEmergencyStop()
 {
     auto current_time = ros::Time::now();
     return (!this->m_last_emergencystop.is_zero() &&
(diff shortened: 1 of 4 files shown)


ros_ws/src/car_control/src/dms_controller.cpp
```diff
@@ -63,6 +60,20 @@ void DMSController::publishDriveMode()
     this->m_drive_mode_publisher.publish(drive_mode_message);
 }
 
+bool DMSController::getEmergencyStop()
+{
+    auto current_time = ros::Time::now();
+    return (!this->m_last_emergencystop.is_zero() &&
+            this->m_last_emergencystop + this->m_emergencystop_expiration_time > current_time);
```
marian42: Please also check if `this->m_last_emergencystop < current_time`. Otherwise, this creates false positives if the simulation is reset.

PR #332, opened 2019-08-25
Added simulation arg to robot xml

### LINKED ISSUE #327: Fix error message in Rviz on the real car
The error message is: No transform from [left_steering] to [laser]

Nils-Jung:
fixes #327 (?)

### DIFF
--- a/ros_ws/launch/car.launch
+++ b/ros_ws/launch/car.launch
@@ -22,6 +22,8 @@
 
     <node name="robot_state_publisher" pkg="robot_state_publisher" type="robot_state_publisher" respawn="false" output="screen"/>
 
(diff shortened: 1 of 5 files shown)


marian42:
This solves the problem in the sense that the error message goes away, but it removes the wheels altogether. I think the ideal solution would be to display information about steering and throttle by rotating the wheels accordingly, but I'm not sure if that's worth the effort.

Nils-Jung:
The main problem is that Rviz expects a /tf transformation from base_link to the wheels and steering but there none since they are exclusive to the simulation. The only way to fix this is to calculate the appropriate transformations from the Focbox data but for me that seems like too much effort for a purely graphical fix. I've now added static transforms so the wheels will be displayed but they won't move or spin.

